PR #2, opened 2023-05-15
Don't cache the published secdb

luhring:


imjasonh (approved):

.github/workflows/build-and-publish-secdb.yaml
```diff
@@ -44,4 +44,9 @@ jobs:
 
       - name: 'Upload the security database to a bucket'
         run: |
-          gcloud --quiet alpha storage cp --recursive ${{ github.workspace }}/security.json gs://wolfi-production-registry-destination/os/
+          # Don't cache the security.json.
+          gcloud --quiet alpha storage cp \
+              --cache-control=no-store \
+              --recursive \
```
imjasonh, line 50:
```suggestion
```

We don't need this since it's only uploading one file, and it's potentially confusing.

imjasonh (approved):

PR #31, opened 2023-06-05
Add CVE-2008-1145 to the nack list for webrick.

dlorenc:


ajayk (approved):

ajayk (changes requested):
is it possible to mention the ruby version and change the wording from older to previous version

kaniini (approved):

PR #42, opened 2023-06-15
nack cve for cue

developer-guy:
It does not seem related with the cue pkg, but PTAL @luhring 🙋‍♂️
> _[CVE-2008-4470](https://github.com/advisories/GHSA-g98p-mfm5-59p2)_


luhring (approved):
Nice!

PR #59, opened 2023-07-01
Add nacked Go advisories to the go-1.21 package.

dlorenc:


dlorenc:
We nacked these in go 1.20 and 1.19. I copied the entries over from 1.20.

luhring (approved):
I wonder if it'd make sense for the advisory data format to allow defining these kinds of advisories once, even though we've split up the Go project by versions into multiple packages

PR #70, opened 2023-07-07
Add advisories for CVE-2023-36617 for uri in ruby

luhring:


imjasonh (approved):

ruby-3.1.advisories.yaml
```diff
@@ -3,97 +3,97 @@ package:
...
 
   CVE-2023-0286:
-    - timestamp: 2023-06-23T03:04:27.00Z
+    - timestamp: 2023-06-23T03:04:27Z
       status: not_affected
       justification: component_not_present
 
   CVE-2023-0401:
-    - timestamp: 2023-06-23T03:04:27.00Z
+    - timestamp: 2023-06-23T03:04:27Z
```
mattmoor, line 96: Why did all of these update?

luhring: I think from an decode/encode cycle. This is something I'd like to standardize soon, perhaps with the advisory schema updates.

luhring: Going to resolve for now to keep this moving, but ping me if you have more thoughts/suggestions!

PR #125, opened 2023-08-09
[gradle-8] CVE-2023-2976 & CVE-2020-8908

stormqueen1990:
Update advisories for CVE-2023-2976 and CVE-2020-8908 in the `gradle-8` package.

### DIFF
--- a/gradle-8.advisories.yaml
+++ b/gradle-8.advisories.yaml
@@ -1,0 +1,10 @@
+package:
+  name: gradle-8
+
+advisories:
+  CVE-2023-2976:
+    - timestamp: 2023-08-08T17:07:36.957515-04:00
+      status: under_investigation
+    - timestamp: 2023-08-10T18:11:57.628638-04:00
+      status: fixed
+      fixed-version: 8.2.1-r1
--- a/gradle-8.advisories.yaml
+++ b/gradle-8.advisories.yaml
@@ -2,6 +2,11 @@ package:
   name: gradle-8
 
 advisories:
+  CVE-2020-8908:
+    - timestamp: 2023-08-10T18:12:46.040472-04:00
+      status: fixed
+      fixed-version: 8.2.1-r1
+
   CVE-2023-2976:
     - timestamp: 2023-08-08T17:07:36.957515-04:00
       status: under_investigation


gradle-8.advisories.yaml
```diff
@@ -0,0 +1,15 @@
...
+    - timestamp: 2023-08-10T16:52:55.95955-04:00
+      status: affected
+      action: 'Temporarily patched in downstream on v8.2.1-r1. Slated for correction in upstream: https://github.com/gradle/gradle/issues/25945'
+
+  CVE-2023-2976:
+    - timestamp: 2023-08-08T17:07:36.957515-04:00
+      status: under_investigation
+    - timestamp: 2023-08-10T15:59:03.284422-04:00
+      status: affected
+      action: 'Temporarily patched in downstream on v8.2.1-r1. Slated for correction in upstream: https://github.com/gradle/gradle/issues/25945'
```
pdeslaur, line 15: We should use the status `fixed` here, with the version `8.2.1-r1`. As an example: https://github.com/wolfi-dev/advisories/blob/main/bind.advisories.yaml#L17-L18

Curious to hear from @luhring whether we should also keep the `affected` event to capture that prose 🤔

pdeslaur: And same for the CVE above!

stormqueen1990: Ooopsies! Updated now!

luhring: >Curious to hear from @luhring whether we should also keep the affected event to capture that prose 🤔

Actually, I might suggest a comment in the Melange config itself (in wolfi-dev/os), to let Wolfi maintainers know that the patch can be removed at a particular point in the future. Example here: https://github.com/wolfi-dev/os/blob/main/zot.yaml#L3. It's always nice to identify when we can stop carrying the debt of patches!

I think it's sufficient to say the vuln has been fixed as of this version (like you've done), working off the assumption that it will stay fixed forever more, either via our patch or upstream incorporation.

stormqueen1990: Oof, forgot the comment. Created https://github.com/wolfi-dev/os/pull/4368 to update it.

PR #136, opened 2023-08-11
node-problem-detector: Advisory for CVE-2023-2727

k4leung4:


pdeslaur (approved):
Go vulndb agrees, it marks the code as not importable! https://github.com/golang/vulndb/blob/master/data/excluded/GO-2023-1891.yaml

PR #145, opened 2023-08-11
NACK CVE-2020-8565 & 2020-8564 for nodetaint

tcnghia:
Both patched in k8s.io/client-go@v0.20.0-alpha.2

See https://github.com/golang/vulndb/blob/1e5f087d110c4ae7fdd0d48fe2d09750f25aef53/data/osv/GO-2021-0066.json#L7 and https://github.com/golang/vulndb/blob/1e5f087d110c4ae7fdd0d48fe2d09750f25aef53/data/osv/GO-2021-0064.json#L5

### DIFF
--- a/nodetaint.advisories.yaml
+++ b/nodetaint.advisories.yaml
@@ -15,14 +15,16 @@ advisories:
       impact: This is a Kubernetes API flaw, but we don't include an API server in this package.
 
   CVE-2020-8564:
-    - timestamp: 2023-08-11T11:38:54.02141-07:00
-      status: fixed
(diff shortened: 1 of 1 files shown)


jspeed-meyers (approved):
Nice NACKs!

luhring:
If we were never affected, let's use `not_affected` here in place of `fixed`

tcnghia:
> If we were never affected, let's use `not_affected` here in place of `fixed`

Done! Thanks

luhring (approved):

PR #156, opened 2023-08-11
NACK CVE-2019-3826 prom-{operator,config-reloader}

tcnghia:
There seems to be a version confusion between the prometheus Go library version (v0.46.0) and their binary version v2.46.0, causing these packages to be flagged with a very old CVE that has since been fixed in v2.7.


pdeslaur (approved):

pdeslaur:
Looks like it's also nacked here: https://github.com/golang/vulndb/blob/master/internal/worker/gen_false_positives.go

PR #173, opened 2023-08-14
Mark CVE-2023-29409 as fixed in go 1.20 and 1.21.

dlorenc:


luhring:
Interesting, Grype shouldn't be surfacing CVE-2023-29409 for Go 1.21 (final release). That looks like a bug on their end

luhring (approved):

PR #176, opened 2023-08-15
kubernetes: clarify impact scope of CVE-2019-11255

luhring:
cc: @nsmith5 — let me know if I got this wrong!

### DIFF
--- a/kubernetes-1.24.advisories.yaml
+++ b/kubernetes-1.24.advisories.yaml
@@ -7,3 +7,6 @@ advisories:
       status: not_affected
       justification: vulnerable_code_not_present
       impact: Exploit requires bundling vulnerable versions of csi-providers with our distribution which we do not do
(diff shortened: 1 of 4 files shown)


nsmith5:
It isn't clear to me that our  kubernetes package is actually affected here. You need to specifically install an external csi driver like aws-ebs-csi-driver and then ship old vulnerable versions of the csi sidecar containers to be affected. Some distributions of kubernetes ship these csi drivers with the control plane but we don't.

I know that means I'm essentially disagreeing with where this CVE was filed (ie. It should be filed against the csi drivers), but I can't wrap my head around how our control plane kubernetes components are vulnerable here

hectorj2f (approved):

luhring:
Thanks @nsmith5! I'll DM you so I can make sure I understand, and we can revert this PR if need be!

PR #177, opened 2023-08-15
Revert "kubernetes: clarify impact scope of CVE-2019-11255"

luhring:
This reverts commit 08f2aa06acd1c24eff7d28e4aea8df1e06b673da.

I had misunderstood the impact of the vulnerability — thanks to @nsmith5 for straightening me out!

### DIFF
--- a/kubernetes-1.24.advisories.yaml
+++ b/kubernetes-1.24.advisories.yaml
@@ -7,6 +7,3 @@ advisories:
       status: not_affected
       justification: vulnerable_code_not_present
       impact: Exploit requires bundling vulnerable versions of csi-providers with our distribution which we do not do
-    - timestamp: 2023-08-14T20:58:00.451222-04:00
-      status: affected
-      action: See https://github.com/kubernetes/kubernetes/issues/85233 for mitigation instructions. (Note that kubectl is unaffected.)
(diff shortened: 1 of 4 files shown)


dlorenc:
Just curious - is revert the right approach here or should we add an additional entry after overriding it as not_affected?

luhring:
> Just curious - is revert the right approach here or should we add an additional entry after overriding it as not_affected?

I'm fine with either approach. I leaned toward revert because of how this entry got added — crossed wires, and a cautionary review from @nsmith5 that was accidentally ignored via auto-merge. It felt like more of a git/GitHub-based mistake to me.

But I'm happy to close this and come back with an append-only approach if you think that's more appropriate!

dlorenc:
> But I'm happy to close this and come back with an append-only approach if you think that's more appropriate!

I don't have any preference - just looking for guidance for future ones!

luhring:
Gotcha! Yeah, I think one heuristic in my head is that our team hadn't reached consensus on my analysis, but it was added to the data _anyway_. As opposed to a normal scenario where collectively we thought `foo` was vulnerable, then we reassessed and realized it's not vulnerable, etc.

I think in general "append-only" is preferred. So for consistency I'll do that here, too.

puerco:
This is an interesting case for the VEX trust model. Once we have this data reflected in VEX documents, should we be able to skip an already issued statement from the VEX statement? Let's dump some ideas the next time we meet @luhring

luhring:
Replacing this with: #178

PR #201, opened 2023-08-27
Add ruby-3.3 advisories seeded from ruby 3.2.

dlorenc:


dlorenc:
Ref https://github.com/wolfi-dev/os/pull/4855/

pdeslaur (approved):

luhring:
🤔  We probably want to use current timestamps here — they're intended to capture when we became aware of each event for each advisory (for each package). Let me know if I'm missing something about this situation

pdeslaur:
> 🤔 We probably want to use current timestamps here — they're intended to capture when we became aware of each event for each advisory (for each package). Let me know if I'm missing something about this situation

True - I didn't think about that, but I agree.

luhring:
I just built and scanned `ruby-3.3` to see what was coming up:

```console
$ w scan packages/aarch64/ruby-3.3-dev-3.3.0_p1-r3.apk
Will process: ruby-3.3-dev-3.3.0_p1-r3.apk
✅ No vulnerabilities found
```

I'll close this one for now but we can revisit/redo it if needed!

PR #203, opened 2023-08-27
Advisories for latest `node-problem-detector` version

pdeslaur:
We picked up version 0.8.14, which fixes a few true positives we've been tracking: wolfi-dev/os#4840

Note that CVE-2020-8565 is now incorrectly reported on our package. [Per k8s](https://github.com/kubernetes/kubernetes/issues/95623), Kubernetes is affected until v1.17.13, and our newer version of the package is now at v1.17.17 (reported as v0.17.17 b/c that's how the library version is setup)

It'd be great to encode more information in our future of advisories.

```
wolfictl scan =(curl -sL https://packages.wolfi.dev/os/x86_64/node-problem-detector-0.8.14-r0.apk)
Will process: zshuRKwOc
└── 📄 /usr/bin/node-problem-detector
        📦 k8s.io/client-go v0.17.17 (go-module)
            Medium CVE-2020-8565 GHSA-8cfg-vx93-jvxw fixed in 0.20.0-alpha.2
```

### DIFF
--- a/node-problem-detector.advisories.yaml
+++ b/node-problem-detector.advisories.yaml
@@ -46,6 +46,9 @@ advisories:
     - timestamp: 2023-08-11T12:45:22.307105-07:00
       status: affected
       action: Pending upstream project to pick up k8s.io/kubernetes 0.19.4+.
+    - timestamp: 2023-08-27T16:51:15.477346-07:00
+      status: fixed
+      fixed-version: 0.8.14-r0
 
   CVE-2021-25735:
     - timestamp: 2023-08-11T11:21:29.54486-07:00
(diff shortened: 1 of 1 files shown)


luhring:
>It'd be great to encode more information in our future of advisories.

What did you have in mind?

pdeslaur:
> What did you have in mind?

I considered adding prose to the `fixed` status to indicate this vulnerability is displayed as a false positive, even after the fix is applied. Maybe it's already implicit in the `fixed` status.

luhring:
> I considered adding prose to the `fixed` status to indicate this vulnerability is displayed as a false positive, even after the fix is applied. Maybe it's already implicit in the `fixed` status.

Yeah, if I'm understanding the situation correctly, I think there's nothing more we can/should do there. This would be a bug in the scanner

PR #210, opened 2023-08-30
thanos: advisories for version streams

nsmith5:


thanos-0.31.advisories.yaml
```diff
@@ -0,0 +1,15 @@
+package:
+  name: thanos
```
pdeslaur, line 2: name needs to be updated, I think!

pdeslaur (approved):

PR #215, opened 2023-09-01
Add nacks for php-calendar.

dlorenc:
We already had these in php, but they apply to the subpackage.


pdeslaur:
Advisories are already in `php`!

PR #223, opened 2023-09-06
Export advisories to GCS

pdeslaur:
Adds a workflow heavily inspired from `build-and-publish-secdb` to export a copy of the advisories to a GCS bucket. This will allow other consumers of the advisories to read the content without needing access to the git repo.

### DIFF
--- a/.github/workflows/build-and-publish-yaml.yaml
+++ b/.github/workflows/build-and-publish-yaml.yaml
@@ -1,0 +1,51 @@
+name: Build and publish YAML
+
+on:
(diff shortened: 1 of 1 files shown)


rawlingsj (dismissed):
I wonder if there's an alternative to using GHA crons for this?

luhring (approved):
Seems like a good stepping stone to me. I'll let someone else merge in case there are more design considerations to work out!

rawlingsj (approved):

PR #245, opened 2023-09-16
Add CVE-2023-32558 as fixed to nodejs-20.

dlorenc:
Grype just started flagging this, but it looks like the CPE bounds are incorrect. The node release notes show this as fixed in 20.6.1, but the NVD has that version as still vulnerable.

https://nodejs.org/en/blog/vulnerability/august-2023-security-releases
https://nvd.nist.gov/vuln/detail/CVE-2023-32558

### DIFF
--- a/nodejs-20.advisories.yaml
+++ b/nodejs-20.advisories.yaml
@@ -102,6 +102,13 @@ advisories:
         data:
           fixed-version: 20.5.1-r0
 
+  - id: CVE-2023-32558
+    events:
+      - timestamp: 2023-09-16T13:47:40Z
+        type: fixed
+        data:
+          fixed-version: 20.6.1-r0
+
   - id: CVE-2023-32559
     events:
       - timestamp: 2023-08-10T12:00:44Z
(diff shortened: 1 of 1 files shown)


nodejs-20.advisories.yaml
```diff
@@ -102,6 +102,13 @@ advisories:
         data:
           fixed-version: 20.5.1-r0
 
+  - id: CVE-2023-32558
+    events:
+      - timestamp: 2023-09-16T13:47:40Z
+        type: fixed
+        data:
+          fixed-version: 20.6.1-r0
```
pdeslaur, line 110: Fixed version should be `20.5.1-r0`.

pdeslaur (commented):
Good find! Looks like NIST added their analysis yesterday and made a mistake.

nodejs-20.advisories.yaml
dlorenc: Good catch!

pdeslaur (approved):

PR #264, opened 2023-09-24
Mark a CVE as fixed in kube-downscaler.

dlorenc:


dlorenc:
Ref: https://github.com/wolfi-dev/os/pull/5893

luhring (approved):

PR #281, opened 2023-10-01
Nack several FPs in gitness.

dlorenc:


gitness.advisories.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+  - id: CVE-2020-13401
+    events:
+      - timestamp: 2023-10-01T13:11:53Z
+        type: false-positive-determination
+        data:
+          type: component-vulnerability-mismatch
+          note: This vulnerability applies to the Docker daemon program, not the docker client libraries.
+
+  - id: GHSA-cf6v-9j57-v6r6
```
pdeslaur, line 23: We could use the CVE ID here: `CVE-2023-3515`

dlorenc: Good catch!

luhring (approved):

PR #295, opened 2023-10-04
Adding additional advisories for keycloak

mamccorm:
As part of preparing our keycloak package for a future image, there are a number of false positives returned in scans, most of these are quite old.

This PR updates the keycloak advisories files with entries for each.


luhring (changes requested):
Holding while we review together 😃

luhring (approved):
Looks great! Thanks!

PR #298, opened 2023-10-05
Add additional advisories for keycloak package

mamccorm:
Adding additional advisories for keycloak package. These where false positives. Linked the relevant information.

### DIFF
--- a/keycloak.advisories.yaml
+++ b/keycloak.advisories.yaml
@@ -108,6 +108,14 @@ advisories:
           type: component-vulnerability-mismatch
           note: CVE is being considered by the community a false positive. See https://github.com/keycloak/keycloak/issues/20226
 
+  - id: CVE-2022-28220
+    events:
+      - timestamp: 2023-10-05T13:50:00Z
+        type: false-positive-determination
+        data:
+          type: component-vulnerability-mismatch
+          note: CVE is being considered by the community a false positive. See https://github.com/keycloak/keycloak/issues/20225
+
   - id: CVE-2022-37832
     events:
       - timestamp: 2023-10-04T19:00:00Z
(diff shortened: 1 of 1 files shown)


keycloak.advisories.yaml
```diff
@@ -116,13 +124,37 @@ advisories:
...
+          type: component-vulnerability-mismatch
+          note: CVE is being considered by the community a false positive. See https://github.com/anchore/grype/issues/1138
+
+  - id: CVE-2023-2976
+    events:
+      - timestamp: 2023-10-05T13:50:00Z
+        type: false-positive-determination
+        data:
+          type: component-vulnerability-mismatch
+          note: CVE already fixed in guava version flagged by scanner (v32.0.0). See https://github.com/google/guava/releases/tag/v32.0.0 and https://github.com/keycloak/keycloak/pull/21544
```
luhring, line 141: Help me understand this one a bit better, I don't think I quite follow. Did the scanner parse the version wrong or something? Why does the scanner think this is a vulnerability, and what's wrong about its logic?

mamccorm: Sure, in this case the scanner is flagging a false positive. For example:

[/usr/share/java/keycloak-22.0.4/lib/lib/main/com.google.guava.guava-32.0.0-jre.jar
│       📦 com.google.guava.guava 32.0.0-jre (java-archive)
│           High CVE-2023-2976](url)

It's correctly picked up that the installed version of guava is 32.0.0, but it's wrongly flagging CVE-2023-2976, which is indeed already fixed in this release.

Perhaps could be worded better but wasn't sure how to describe and keep high level

keycloak.advisories.yaml
```diff
@@ -116,13 +124,37 @@ advisories:
...
+
   - id: CVE-2023-0264
     events:
       - timestamp: 2023-04-30T11:15:20Z
         type: false-positive-determination
         data:
-          type: vulnerable-code-version-not-used
-          note: The CVE was fixed in a version of keycloak prior to our packaging. See https://github.com/keycloak/keycloak/issues/21079
+          type: component-vulnerability-mismatch
+          note: CVE was already fixed in v21.0.1 and later. See https://github.com/advisories/GHSA-9g98-5mj6-f9mv
```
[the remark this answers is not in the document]
mamccorm: This was a mistake in my part in my last commit. Spotted this was not pointing to the correct CVE so was correcring as part of this batch of updates.

If theres a different process to follow for updates though i'm happy to follow, ill drop you a 1-1 on this one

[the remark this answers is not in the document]
luhring: Ah I see now. Good catch and thanks for reverting this back. The "wrong version" FP type still sounds correct, though, is that right? (The original value was `vulnerable-code-version-not-used` here)

keycloak.advisories.yaml
luhring: I'd like to understand what's happening here to make the scanner flag the vulnerability. Could it be the version suffix of `-jre`? (Somehow the scanner thinks it's doing the right thing, so I just want to understand why 😄 )

mamccorm: Happy to file a grype ticket, but if theres anything we can do on our side as well lets sync up tomorrow

mamccorm: Hey @luhring, heres the grype ticket:
- https://github.com/anchore/grype/issues/1543

Let me know if you still want to take a look further but there should be plenty of details in there for the grype team.

luhring (approved):
Thanks @mamccorm!

PR #304, opened 2023-10-09
Added advisory for CVE-2023-4586

mamccorm:
Added detection advisory with summary and links for CVE-2023-4586, which is detected while scanning Keycloak.

### DIFF
--- a/keycloak.advisories.yaml
+++ b/keycloak.advisories.yaml
@@ -176,14 +176,10 @@ advisories:
   - id: CVE-2023-4586
     events:
       - timestamp: 2023-10-09T10:53:00Z
-        type: detection
+        type: true-positive-determination
         data:
-          type: manual
           note: |
-            This CVE exists in the latest release (v4) of netty-handler, which is a transitive dependency
-            of Quarkus, used by Keycloak. The netty-handler team state this will not be present in a newer
-            release (v5.x), but this is not public and has been in developent for quite some time. It remains
-            to be seen whether the latest released version will receive a patch. See also:
-            https://github.com/opensearch-project/security/issues/3477
-            https://github.com/keycloak/keycloak/blob/21a23ace1d0aaada4825f4627b9bc3835d190ff3/.github/snyk/.snyk#L21
-            https://security.snyk.io/vuln/SNYK-JAVA-IONETTY-1042268
+            This CVE exists in the latest release (v4) of netty-handler, which has yet to receive a patch.
+            There is discussion this would be mitigated in an upcoming v5 of netty, but this is yet to be released and there is no ETA.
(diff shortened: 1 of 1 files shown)


keycloak.advisories.yaml
```diff
@@ -172,3 +172,18 @@ advisories:
...
+
+  - id: CVE-2023-4586
+    events:
+      - timestamp: 2023-10-09T10:53:00Z
+        type: detection
+        data:
+          type: manual
+          note: |
+            This CVE exists in the latest release (v4) of netty-handler, which is a transitive dependency
+            of Quarkus, used by Keycloak. The netty-handler team state this will not be present in a newer
```
luhring, line 184: In general, let's try to avoid adding line breaks mid-sentence in this data, since we're not sure how the data will ultimately be rendered downstream, and we don't have an established convention for this data that would make these line breaks nonissues for downstream consumers (such as "treat this data as markdown" or something).

keycloak.advisories.yaml
```diff
@@ -172,3 +172,18 @@ advisories:
...
           type: vulnerability-record-analysis-contested
           note: CVE is being considered by the community a false positive. See https://github.com/FasterXML/jackson-databind/issues/3972 and https://github.com/anchore/grype/issues/1386
+
+  - id: CVE-2023-4586
+    events:
+      - timestamp: 2023-10-09T10:53:00Z
+        type: detection
+        data:
+          type: manual
+          note: |
```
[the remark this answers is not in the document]
mamccorm: marking this as resolved as the bug has been fixed and i've also reverted from using detection, as per below feedback

keycloak.advisories.yaml
mamccorm: Sure, good to know and noted - updated accordingly and ready for another look - @luhring

keycloak.advisories.yaml
```diff
@@ -172,3 +172,14 @@ advisories:
...
+
+  - id: CVE-2023-4586
+    events:
+      - timestamp: 2023-10-09T10:53:00Z
+        type: true-positive-determination
+        data:
+          note: |
+            This CVE exists in the latest release (v4) of netty-handler, which has yet to receive a patch.
+            There is discussion this would be mitigated in an upcoming v5 of netty, but this is yet to be released and there is no ETA.
+            nettty-handler is a dependency of Quarkus, which is used by Keycloak. For more detail:
```
luhring, line 184:
```suggestion
            netty-handler is a dependency of Quarkus, which is used by Keycloak. For more detail:
```

luhring (approved):
Thanks @mamccorm! 🙏

PR #366, opened 2023-10-19
advisory for rabbitmq operator

k4leung4:
https://github.com/wolfi-dev/os/pull/7087
https://github.com/wolfi-dev/os/pull/7090


cpanato (approved):

luhring (changes requested):
Actually we don't need to log **fixes** for net new packages, if we've never shipped "affected" versions of them!

k4leung4:
> Actually we don't need to log **fixes** for net new packages, if we've never shipped "affected" versions of them!

thanks, til

PR #386, opened 2023-10-24
Add CVE-2023-5363 for openssl as fixed.

dlorenc:


dlorenc:
https://github.com/wolfi-dev/os/pull/7314

openssl.advisories.yaml
```diff
@@ -199,3 +199,10 @@ advisories:
         data:
           type: vulnerable-code-not-included-in-package
           note: Component is only vulnerable when running under Windows.
+
+  - id: CVE-2023-5363
+    events:
+      - timestamp: 2023-10-24T15:21:00Z
+        type: fixed
+        data:
+          fixed-version: 3.1.4-r0
```
cpanato, line 208: my understanding was if is a updated package (epoch 0) we dont need to add the adv, am I wrong? cc @luhring

cpanato (approved):

openssl.advisories.yaml
luhring: We should still note the fix, even if it was provided by upstream

luhring (approved):

PR #402, opened 2023-10-26
add entry for terraform regarding GHSA-m425-mq94-257g cve

cpanato:
related to https://github.com/wolfi-dev/os/pull/7441

### DIFF
--- a/terraform.advisories.yaml
+++ b/terraform.advisories.yaml
@@ -24,3 +24,10 @@ advisories:
         type: fixed
         data:
           fixed-version: 1.5.5-r6
+
+  - id: GHSA-m425-mq94-257g
+    events:
+      - timestamp: 2023-10-26T11:40:34Z
+        type: fixed
+        data:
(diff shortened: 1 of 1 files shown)


terraform.advisories.yaml
```diff
@@ -24,3 +24,10 @@ advisories:
         type: fixed
         data:
           fixed-version: 1.5.5-r6
+
+  - id: GHSA-m425-mq94-257g
```
hectorj2f, line 28: Don't we have the normalized CVE ID ? So we can use that one instead, and set this one as the alias.

cpanato: looks like they dont have direct connection

hectorj2f (approved):

terraform.advisories.yaml
hectorj2f: Okay. It can happen sometimes.

luhring (approved):

PR #412, opened 2023-10-27
Advisories for trino

pdeslaur:
So.... I ended up marking all the `trino` vulnerabilities as true-positive. Arguably CVE-2023-44981 is LIKELY a false-positive, but I can verify that the offending class is included in the package. At that point, I feel more comfortable marking it as a true positive.


k4leung4 (approved):

luhring (approved):
I love the thoroughness here ❤️

PR #432, opened 2023-11-01
reports for npm package CVEs

developer-guy:


luhring:
This is looking good. For the notes for `type: pending-upstream-fix`, it'd be great if we can explain why we're not fixing it ourselves and why upstream needs to be the ones to do it

kubeflow-pipelines.advisories.yaml
```diff
@@ -192,11 +238,72 @@ advisories:
...
+        type: pending-upstream-fix
+        data:
+          note: crypto-js should be upgraded to 4.2.0
+
+  - id: GHSA-2jcg-qqmg-46q6
+    events:
+      - timestamp: 2023-11-01T07:13:06Z
+        type: pending-upstream-fix
+        data:
+          note: monorepo-symlink-test should be upgraded
```
julienv3, line 288: This is a false positive;
https://github.com/browserify/resolve/issues/288

kubeflow-pipelines.advisories.yaml
```diff
@@ -43,25 +43,26 @@ advisories:
...
           note: Waiting for upstream release to pick up one of k8s.io/kubernetes 0.17.0+.
 
-  - id: CVE-2019-11253
-    aliases:
-      - GHSA-pmqp-h87c-mr78
+  - id: CVE-2020-15168
     events:
-      - timestamp: 2023-09-30T14:48:25Z
-        type: false-positive-determination
+      - timestamp: 2023-11-01T07:08:15Z
```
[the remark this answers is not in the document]
developer-guy: I didn't mean to delete anything this was a mistake while resolving conflicts by me sorry about that

kubeflow-pipelines.advisories.yaml
```diff
@@ -43,25 +43,26 @@ advisories:
         data:
           note: Waiting for upstream release to pick up one of k8s.io/kubernetes 0.17.0+.
 
-  - id: CVE-2019-11253
-    aliases:
-      - GHSA-pmqp-h87c-mr78
+  - id: CVE-2020-15168
```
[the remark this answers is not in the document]
developer-guy: afaik, this aliases won't add automatically at the time of writing this, we add them manually.

kubeflow-pipelines.advisories.yaml
```diff
@@ -43,25 +43,26 @@ advisories:
...
-          type: vulnerable-code-not-included-in-package
-          note: Go vulndb has marked this code NOT_IMPORTABLE.
+          note: date-and-time should be upgraded 0.14.2
+
+  - id: CVE-2020-8237
+    events:
+      - timestamp: 2023-11-01T06:55:52Z
+        type: pending-upstream-fix
+        data:
+          note: json-bigint should be upgraded 1.0.0
```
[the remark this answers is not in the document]
developer-guy: amazing!

kubeflow-pipelines.advisories.yaml
```diff
@@ -192,11 +288,77 @@ advisories:
...
+          note: cannot fix the CVE because there were no patched versions since the 'request' package is no longer supported by the maintainer.
+
+  - id: CVE-2023-34104
+    aliases:
+      - GHSA-6w63-h3fj-q4vw
+    events:
+      - timestamp: 2023-11-01T06:56:37Z
+        type: pending-upstream-fix
+        data:
+          note: fast-xml-parser should be upgraded to 4.2.4
```
mamccorm, line 314: This comment will need updated as well

kubeflow-pipelines.advisories.yaml
```diff
@@ -192,11 +288,77 @@ advisories:
...
+
+  - id: CVE-2023-43804
+    events:
+      - timestamp: 2023-11-01T11:36:40Z
+        type: pending-upstream-fix
+        data:
+          note: urllib3 should be upgraded to 1.26.17
```
mamccorm, line 321: This comment will need updated as well

kubeflow-pipelines.advisories.yaml
```diff
@@ -192,11 +288,77 @@ advisories:
...
+
+  - id: CVE-2023-45803
+    events:
+      - timestamp: 2023-11-01T11:36:15Z
+        type: pending-upstream-fix
+        data:
+          note: urllib3 should be upgraded to 1.26.18
```
mamccorm, line 328: This comment will need updated as well

kubeflow-pipelines.advisories.yaml
```diff
@@ -192,11 +288,77 @@ advisories:
...
+
+  - id: CVE-2023-46233
+    events:
+      - timestamp: 2023-11-01T06:38:31Z
+        type: pending-upstream-fix
+        data:
+          note: crypto-js should be upgraded to 4.2.0
```
mamccorm, line 335: This comment will need updated as well

kubeflow-pipelines.advisories.yaml
```diff
@@ -192,11 +288,77 @@ advisories:
...
+        type: pending-upstream-fix
+        data:
+          note: '''node-forge'' is a transitive dependency of ''google-p12-pem'', which is pinned to an old release. Attempts to patch have been unsuccessful, as the new releases introduced a lot of changes. Waiting on upstream to address.'
+
+  - id: GHSA-x3cc-x39p-42qx
+    events:
+      - timestamp: 2023-11-01T07:03:32Z
+        type: pending-upstream-fix
+        data:
+          note: fast-xml-parser should be upgraded to 4.1.2
```
mamccorm, line 364: This comment will need updated as well

kubeflow-pipelines.advisories.yaml
developer-guy: fast-xml-parser this one and this crypto-js are both main dependencies. So, I tried to bump their version through the `overrides` section in the package.json but it didn't work so I left the note as `fast-xml-parser should be upgraded to <version>`.

kubeflow-pipelines.advisories.yaml
developer-guy: fast-xml-parser this one and this crypto-js are both main dependencies. So, I tried to bump their version through the `overrides` section in the package.json but it didn't work so I left the note as `fast-xml-parser should be upgraded to <version>`.

kubeflow-pipelines.advisories.yaml
developer-guy: fast-xml-parser this one and this crypto-js are both main dependencies. So, I tried to bump their version through the `overrides` section in the package.json but it didn't work so I left the note as `fast-xml-parser should be upgraded to <version>`.

kubeflow-pipelines.advisories.yaml
developer-guy: fast-xml-parser this one and this crypto-js are both main dependencies. So, I tried to bump their version through the `overrides` section in the package.json but it didn't work so I left the note as `fast-xml-parser should be upgraded to <version>`.

kubeflow-pipelines.advisories.yaml
developer-guy: same.

kubeflow-pipelines.advisories.yaml
mamccorm: @developer-guy  we're still going to need a better description than what we have now. Think of it from the customers perspective, they would look at this feed and see that we've marked as pending-upstream-fix, and we've told them something should be upgraded.

Thats not clear at all so we'll need to re-work the note

mamccorm: This one looks like we have the ability to patch. Going from v4.2.5 to v4.3.2 of the fast-xml-parser plugin seems reasonable. I pulled down the source and i was able to patch the `frontend/server/package-lock.json` and compile successfully.

Also it does not look like the upstream [package-lock.json](https://github.com/kubeflow/pipelines/commits/master/frontend/package-lock.json) is updated very often. And when it is updated it looks like individual line changes mostly vs a re-generation of the whole file. I think we should be okay to patch this? 

@julienv3 thoughts?

developer-guy: I'd like to highlight some point tough because the CVE says you should upgrade `fast-xml-parser` to 4.2.4 since you were using the older version `3.21.1` but when I had a look at the source code of the frontend project, there is no version like `3.21.1` for `fast-xml-parser`, so, we should understand where it actually comes from first. Because according to the source code we seem using `4.2.5` as dep of `@aws-sdk/client-sts`, [here](https://github.com/kubeflow/pipelines/blob/d9c3b86adc4dc03beb623dc6ec8230347051200c/frontend/server/package-lock.json#L308).

julienv3: The changelog https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/CHANGELOG.md doesn't mention breaking changes, you can try it out and see if the frontend still loads, although again this is hard to validate unless we have our own full test setup - it's a runtime dependency, it might be used in specific conditions, and different input could cause different behavior.

kubeflow-pipelines.advisories.yaml
developer-guy: I fixed that issue in the package, so we don't this one anymore.

kubeflow-pipelines.advisories.yaml
developer-guy: I fixed that in the package so we don't need this anymore.

kubeflow-pipelines.advisories.yaml
```diff
@@ -200,3 +210,20 @@ advisories:
         data:
           type: vulnerable-code-not-included-in-package
           note: Vulnerable code is part of the Kubernetes ServiceAccount admission plugin which is not importable and also code was marked NOT_IMPORTABLE in Golang vulndb
+
+  - id: CVE-2023-28155
+    aliases:
+      - GHSA-p8p7-x288-28g6
+    events:
+      - timestamp: 2023-11-01T07:12:35Z
+        type: fix-not-planned
```
pdeslaur, line 219: `fix-not-planned` implies that we don't ever expect to see this vulnerability to go away. Given this package isn't pinned to an older version, I'd recommend using the status `pending-upstream-fix`. Upstream in this case refers to the `kubeflow-pipelines` project. That project could eventually migrate away from needing the `request` package.

developer-guy: okay that makes sense, thanks @pdeslaur 🤝

luhring (approved):

PR #454, opened 2023-11-03
fix: GHSA-m425-mq94-257g

hectorj2f:
related to https://github.com/wolfi-dev/os/pull/7955/files


tcnghia (dismissed):

ko.advisories.yaml
```diff
@@ -49,6 +49,13 @@ advisories:
         data:
           fixed-version: 0.13.0-r3
 
+  - id: CVE-2023-44487
```
luhring, line 52: Can we add the GHSA alias too?

hectorj2f: sure

luhring (approved):

PR #456, opened 2023-11-03
fix: GHSA-m425-mq94-257g

hectorj2f:
related to https://github.com/wolfi-dev/os/pull/7953/files


luhring (commented):
Let's add the alias too

luhring (approved):

PR #462, opened 2023-11-04
nvidia: GHSA-m425-mq94-257g

hectorj2f:
related to https://github.com/wolfi-dev/os/pull/7970/files

### DIFF
--- a/nvidia-device-plugin.advisories.yaml
+++ b/nvidia-device-plugin.advisories.yaml
@@ -1,4 +1,4 @@
-schema-version: "2"
+schema-version: 2.0.1
 
 package:
   name: nvidia-device-plugin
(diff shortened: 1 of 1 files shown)


nvidia-device-plugin.advisories.yaml
```diff
@@ -19,8 +19,14 @@ advisories:
...
 
   - id: CVE-2023-44487
+    aliases:
+      - GHSA-m425-mq94-257g
     events:
       - timestamp: 2023-10-16T18:53:58Z
         type: fixed
         data:
           fixed-version: 0.14.1-r5
+      - timestamp: 2023-11-04T13:00:09Z
```
cpanato, line 29: this is okay to do? cc @luhring

hectorj2f: This is the output of `wolfictl adv update`. It has been done in other advisory files.

luhring: What's the question exactly? "Is it okay to have more than one event per advisory?"?

pdeslaur: I think it's okay! We detected the same CVE twice for different components. It's a bit funky. We automate detection events, this will make more sense imo

luhring (approved):

nvidia-device-plugin.advisories.yaml
luhring: Agreed! The latest of the `fixed` events will "win out" when it comes to the secdb, which is good in terms of notifying affected users of our package.

PR #467, opened 2023-11-06
Configure GitHub action to ingest advisories in BigQuery

pdeslaur:
This will enable real-time updates to the BigQuery dataset.


luhring (approved):
😍

PR #471, opened 2023-11-06
grype fix GHSA-m425-mq94-257g

hectorj2f:
related to https://github.com/wolfi-dev/os/pull/8069


grype.advisories.yaml
```diff
@@ -4,6 +4,13 @@ package:
   name: grype
 
 advisories:
+  - id: CVE-2023-44487
```
luhring, line 7: Can you include the alias, too?

luhring (approved):

PR #476, opened 2023-11-06
helm: add advisories for GHSA-m425-mq94-257g

ajayk:


pdeslaur (approved):
That's a fun one! It'll be cool when we include detection information in advisories, that type of situation will make more sense

PR #492, opened 2023-11-09
metrics-server: add advisories for GHSA-m425-mq94-257g

ajayk:


metrics-server.advisories.yaml
```diff
@@ -128,3 +128,10 @@ advisories:
         data:
           type: vulnerable-code-not-included-in-package
           note: Only affects Windows
+
+  - id: GHSA-m425-mq94-257g
```
luhring, line 132: Can we use the CVE for the ID, and add this GHSA as an alias?

luhring (approved):

PR #497, opened 2023-11-12
php: False positive advisory for CVE-2022-4900

pdeslaur:
That's a fairly new CVE - but it looks like it was fixed a little while back!


luhring (approved):
Nice! ✨

PR #503, opened 2023-11-13
metrics-server: CVE-2023-47108

hectorj2f:
This CVE cannot be fixed without applying major changes to the current source code. In addition, it seems like there isn't any new release since July.

### DIFF
--- a/metrics-server.advisories.yaml
+++ b/metrics-server.advisories.yaml
@@ -133,3 +133,12 @@ advisories:
         data:
           type: vulnerable-code-not-included-in-package
           note: Only affects Windows
+
+  - id: CVE-2023-47108
+    aliases:
+      - GHSA-8pgv-569h-w5rw
+    events:
+      - timestamp: 2023-11-13T18:02:45Z
+        type: pending-upstream-fix
+        data:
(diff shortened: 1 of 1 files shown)


metrics-server.advisories.yaml
```diff
@@ -133,3 +133,12 @@ advisories:
...
           note: Only affects Windows
+
+  - id: CVE-2023-47108
+    aliases:
+      - GHSA-8pgv-569h-w5rw
+    events:
+      - timestamp: 2023-11-13T18:02:45Z
+        type: pending-upstream-fix
+        data:
+          note: Pending upstream fix, this fix will require some code changes.
```
pdeslaur, line 144: Would it be possible to include more information on what we're waiting for? The metrics-server repo has had several dependency updates since the last release, but we're still waiting for upstream to cut a new version. 

Let's include that information - the more the better!

hectorj2f: Sure, I can add more details to the note.

hectorj2f: I added a detailed description, not sure if this is what you wanted.

pdeslaur (approved):

PR #504, opened 2023-11-13
add GHSA-m425-mq94-257g/GHSA-qppj-fm5r-hxr3 and GHSA-jq35-85cj-fj4p for kubeflow-katib

cpanato:
related to https://github.com/wolfi-dev/os/pull/8429

### DIFF
--- a/kubeflow-katib.advisories.yaml
+++ b/kubeflow-katib.advisories.yaml
@@ -22,6 +22,16 @@ advisories:
         data:
           fixed-version: 0.15.0-r7
 
+  - id: CVE-2023-44487
+    aliases:
+      - GHSA-m425-mq94-257g
+      - GHSA-qppj-fm5r-hxr3
+    events:
+      - timestamp: 2023-11-13T17:06:53Z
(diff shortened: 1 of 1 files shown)


kubeflow-katib.advisories.yaml
```diff
@@ -22,6 +22,16 @@ advisories:
         data:
           fixed-version: 0.15.0-r7
 
+  - id: CVE-2023-44487
+    aliases:
+      - GHSA-m425-mq94-257g
```
pdeslaur, line 27: Are you sure this alias is correct? I see it reported on 2 different lines in `wolfictl scan`

pdeslaur: CVE-2023-44487 reports as a false-positive by `govulncheck` - we should record the advisory as such (even if we bumped the version of google.golang.org/grpc

cpanato: for the grpc issue this is what we did for the others as well

luhring (approved):

PR #513, opened 2023-11-15
trino: advisory for GHSA-2cqf-6xv9-f22w

pdeslaur:
Further analysis of the Trino source code convinces me that Trino only relies on the client libraries and not running the server. Thus, this CVE is a false positive.


luhring (dismissed):
(holding per offline discussion: CVE-2023-31418/GHSA-2cqf-6xv9-f22w still showing up in package scan)

pdeslaur:
Finally got back to this. After further analysis, this cannot be bumped easily upstream. I got through some of the required code changes, but tl;dr, silencing the vulnerability involves migrating Trino from the elastic search library v6 to v7+. There are a number of breaking changes in the underlying network library, that makes it scary to patch.

Further analysis of the Trino source code convinces me that Trino only relies on the client libraries and not running the server. Thus, this CVE is a false positive.

cpanato (approved):

PR #529, opened 2023-11-16
fix: duplicate advisories for CVE-2023-44487

luhring:
Caught by new validation in `wolfictl`! 🎉


luhring:
🚨  This is expected to fail diff validation since events are being removed.

ajayk (approved):

luhring:
Bypassing branch protections per the above explanation

PR #540, opened 2023-11-20
add CVE-2023-39318 / CVE-2023-39319 / CVE-2023-44487 / CVE-2023-39323 for go-fips-1.19

cpanato:


luhring:
I don't think we need these any more

PR #543, opened 2023-11-21
fix: correct nonexistent fixed-versions

luhring:
These cases of `fixed-version` references to versions that don't exist in the APKINDEX were found by in-progress validation logic in `wolfictl`. In some cases, the versions found have been withdrawn. In other cases, it looks like `-r0` was assumed to exist but it didn't at time of entry.

cc: @jonjohnsonjr 

**Note:** This is expected to fail validation because it's modifying existing event data.


luhring:
>Note: This is expected to fail validation because it's modifying existing event data.

🤔

jonjohnsonjr (approved):

PR #583, opened 2023-12-03
Nack three CVEs in busybox.

dlorenc:
These all require FEATURE_CLEANUP to be set at compile time, which we don't set. Ref: https://github.com/wolfi-dev/os/blob/main/busybox/busyboxconfig#L36

Busybox upstream does not depend on free() logic to clean up, it uses exit() becuase it is designed to be short-lived. The non-default FEATURE_CLEAN_UP setting does trigger free() calls, but we don't compile with that.

### DIFF
--- a/busybox.advisories.yaml
+++ b/busybox.advisories.yaml
@@ -1,4 +1,4 @@
-schema-version: "2"
+schema-version: 2.0.1
 
 package:
   name: busybox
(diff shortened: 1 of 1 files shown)


kaniini:
Looks good to me!

luhring:
Thanks @dlorenc!

PR #603, opened 2023-12-11
adds temporal advisory for otel CVE

mritunjaysharma394:
This CVE cannot be fixed without applying major changes to the current source code. In addition, it seems like they are not visible even in the new possible release https://github.com/temporalio/temporal/blob/v1.23.0-rc0/

### DIFF
--- a/temporal-server.advisories.yaml
+++ b/temporal-server.advisories.yaml
@@ -1,4 +1,4 @@
-schema-version: 2.0.1
+schema-version: 2.0.2
 
 package:
   name: temporal-server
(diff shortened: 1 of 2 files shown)


pdeslaur:
Do you have a bit more information about what the failure is when bumping otel? Bumping from v0.42.0 to v0.46.0 is something we've managed in the past

mritunjaysharma394:
Yes, here's my conversation with @cpanato and @hectorj2f in the past where I faced similar error and even this time I was getting the same error, I tried a lot of things we did in past but it didn't help:
https://chainguard-dev.slack.com/archives/C0322APD4LU/p1700223084301419

shivanshu1333:
this is also needed for https://github.com/chainguard-dev/enterprise-packages/pull/1172

luhring (approved):

PR #611, opened 2023-12-14
Creating new grafana advisory for CVE-2019-3826, GHSA-3m87-5598-2v4f

mamccorm:
Creating new grafana advisory for: CVE-2019-3826, GHSA-3m87-5598-2v4f.

### DIFF
--- a/grafana.advisories.yaml
+++ b/grafana.advisories.yaml
@@ -23,3 +23,13 @@ advisories:
         data:
           type: vulnerable-code-not-included-in-package
           note: Only affects Windows
+
+  - id: CVE-2019-3826
+    aliases:
+      - GHSA-3m87-5598-2v4f
+    events:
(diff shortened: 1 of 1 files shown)


rawlingsj (approved):

grafana.advisories.yaml
```diff
@@ -23,3 +23,13 @@ advisories:
...
           note: Only affects Windows
+
+  - id: CVE-2019-3826
+    aliases:
+      - GHSA-3m87-5598-2v4f
+    events:
+      - timestamp: 2023-12-14T10:32:41Z
+        type: false-positive-determination
+        data:
+          type: component-vulnerability-mismatch
```
luhring, line 34: It sounds like this is the right component but the wrong version — is that right?

If so let's try to use `vulnerable-code-version-not-used` instead for cases like this, please!

cc: @mamccorm @rawlingsj

PR #725, opened 2024-01-04
Clean up some excess permissions.

mattmoor:
I believe the digestabot one is leftover from before we used the PAT, and I don't believe we actually need `packages: write` for the other workflows, but I'm confirming that.


luhring (approved):

mattmoor:
Holler if anything breaks, but I suspect the `packages: write` is bad copy/pasta.

PR #754, opened 2024-01-09
Adding CVEs for bluez-dbg

chainguardian:
Adding Advisory CVE-2020-24490 for bluez-dbg


cpanato:
added in https://github.com/wolfi-dev/advisories/pull/750

PR #771, opened 2024-01-10
fix(ci): use pull_request instead of pull_request_target

luhring:
This appears to be connected to the faulty validation in CI, because of the git references used in the pipeline when `pull_request_target` is used, which seems to result in CI validating the `main` branch.

For more information, see: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

cc: @jonjohnsonjr @cpanato

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -1,7 +1,7 @@
 name: Validate
 
 on:
-  pull_request_target:
+  pull_request:
     branches: ['main']
   push:
     branches:
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -34,8 +34,6 @@ jobs:
     - name: Validate advisory data
       id: validate
       working-directory: ./wolfi-advisories
-      env:
-        GITHUB_TOKEN: ${{ secrets.WOLFI_BOT_PAT_ADVISORIES }}
       run: |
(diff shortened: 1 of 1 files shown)


k4leung4:
`pull_request_target` is used because `pull_request` does not have access to secrets from forked repos

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges

```
With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. The GITHUB_TOKEN has read-only permissions in pull requests from forked repositories. For more information, see "[Automatic token authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)."
```

luhring:
> `pull_request_target` is used because `pull_request` does not have access to secrets from forked repos

Does this workflow need secrets?

k4leung4:
i see this 

```
GITHUB_TOKEN: ${{ secrets.WOLFI_BOT_PAT_ADVISORIES }}
```

https://github.com/wolfi-dev/advisories/blob/main/.github/workflows/validate.yaml#L38

though it is not clear to me if it actually needs it

luhring:
Good eye! I just deleted that to see what happens 😄 

I'm up for trying this if you are. The one thing I know is that we can't leave the workflow the way it is, since it's messing up CI's ability to validate the advisories.

k4leung4 (approved):
lets try it

PR #807, opened 2024-01-11
cassandra-reaper

developer-guy:
- https://github.com/wolfi-dev/os/pull/10893

/cc @Dentrax

### DIFF
--- a/cassandra-reaper.advisories.yaml
+++ b/cassandra-reaper.advisories.yaml
@@ -20,6 +20,10 @@ advisories:
             componentType: java-archive
             componentLocation: /usr/local/lib/cassandra-reaper.jar
             scanner: grype
+      - timestamp: 2024-01-18T10:02:13Z
+        type: pending-upstream-fix
+        data:
+          note: To be able to fix this CVE, we need to upgrade the "org.apache.cassandra:cassandra-all" dependency version from `2.2.12` which is the version project is currently using to at least `4.0.10` which is the version we should bump to to fix the CVEs but we can't because the build was failed due to compilation errors.
 
   - id: CVE-2020-8908
     aliases:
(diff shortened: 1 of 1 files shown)


cassandra-reaper.advisories.yaml
```diff
@@ -0,0 +1,32 @@
...
+          note: To be able to fix this CVE, we need to upgrade the "org.apache.cassandra:cassandra-all" dependency version from `2.2.12` which is the version project is currently using to at least `4.0.10` which is the version we should bump to to fix the CVEs but we can't because the build was failed due to compilation errors.
+
+  - id: CVE-2020-8908
+    aliases:
+      - GHSA-5mg8-w23w-74h3
+    events:
+      - timestamp: 2024-01-11T09:54:49Z
+        type: pending-upstream-fix
+        data:
+          note: Pending upstream fix, this fix will require some code changes since when we upgrade the "com.google.guava:guava" dependency version from `24.1.1-jre` which is the version project is currently using to `32.0.0-android` which is the version we should bump to to fix the CVEs but we can't because the build was failed due to compilation errors.
```
stormqueen1990, line 23: Perhaps we can just mention the version here without the variant?

```suggestion
          note: Pending upstream fix, this fix will require some code changes since when we upgrade the "com.google.guava:guava" dependency version from 24.1.1 which is the version project is currently using to 32.0.0 which is the version we should bump to to fix the CVEs but we can't because the build was failed due to compilation errors.
```

cassandra-reaper.advisories.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+  - id: CVE-2023-2976
+    aliases:
+      - GHSA-7g45-4rm6-3mm3
+    events:
+      - timestamp: 2024-01-11T09:59:33Z
+        type: pending-upstream-fix
+        data:
+          note: Pending upstream fix, this fix will require some code changes since when we upgrade the "com.google.guava:guava" dependency version from `24.1.1-jre` which is the version project is currently using to `32.0.0-android` which is the version we should upgrade to fix the CVEs but we can't because the build was failed due to compilation errors.
```
stormqueen1990, line 32:
```suggestion
          note: Pending upstream fix, this fix will require some code changes since when we upgrade the "com.google.guava:guava" dependency version from 24.1.1 which is the version project is currently using to 32.0.0 which is the version we should upgrade to fix the CVEs but we can't because the build was failed due to compilation errors.
```

stormqueen1990 (commented):
LGTM!

luhring:
Should only need a rebase and current timestamps on the new entries, and this looks good to go!

luhring (approved):

luhring (approved):
@developer-guy thanks for your patience on this one!

PR #840, opened 2024-01-13
metric-collector-for-apache-cassandra

developer-guy:


metric-collector-for-apache-cassandra.advisories.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+advisories:
+  - id: CVE-2022-1471
+    aliases:
+      - GHSA-mjmj-j48q-9wg2
+    events:
+      - timestamp: 2024-01-13T08:38:03Z
+        type: pending-upstream-fix
+        data:
+          note: Pending upstream fix, this fix will require some code change
```
luhring, line 14: This doesn't explain why we can't fix it ourselves. (Also, no need to say "pending upstream fix" in the note text if "pending-upstream-fix" is already the event type!)

luhring (changes requested):

metric-collector-for-apache-cassandra.advisories.yaml
developer-guy: I was searching for the same CVE within the previous CVE entries, I used that one because the similar issue, sorry about that. I can give more detail about why the build fails due to upgrade.

metric-collector-for-apache-cassandra.advisories.yaml
```diff
@@ -0,0 +1,15 @@
...
+advisories:
+  - id: CVE-2022-1471
+    aliases:
+      - GHSA-mjmj-j48q-9wg2
+    events:
+      - timestamp: 2024-01-13T08:38:03Z
+        type: pending-upstream-fix
+        data:
+          note: |
+            To fix the CVE, we have to upgrade 'snakeyaml' to '2.0' or later but this fix will require some code change since the upgrade cause to build fail due to compilation errors like: "/src/main/java/com/datastax/mcac/ConfigurationLoader.java:[106,19] incompatible types: java.lang.Class<capture#1 of ?> cannot be converted to org.yaml.snakeyaml.LoaderOptions"
```
luhring, line 15: This is awesome, thanks @developer-guy!

LGTM, you'll just need to rebase I think

luhring (dismissed):

luhring (approved):

PR #845, opened 2024-01-14
skaffold: add adv data

hectorj2f:


cpanato (approved):

skaffold.advisories.yaml
```diff
@@ -95,6 +99,10 @@ advisories:
             componentType: go-module
             componentLocation: /usr/bin/skaffold
             scanner: grype
+      - timestamp: 2024-01-14T11:45:35Z
+        type: pending-upstream-fix
+        data:
+          note: This vulnerability requires changes in the upstream code to support a containerd version with the fixed vulnerability.
```
luhring, line 105: Just a thought for next time — it could be useful to provide a bit more context about what the problem was in this note. As it is now, this note expresses little more than what's already evident in the advisory data — i.e. that the problem is in `containerd` and that we're waiting on an upstream fix.

hectorj2f: @luhring What would you like to see under these circumstances ? If we add the failing code or the file or function call that would make this description more clear. 

Could you also share an example of a descriptive note ?

luhring: >If we add the failing code or the file or function call that would make this description more clear.

Yeah that's perfect! Just enough to help the next person that comes along know what you tried and what problem to look for.

Here are some ones that give good context:

- https://github.com/wolfi-dev/advisories/blob/main/temporal-server.advisories.yaml#L36-L39
- https://github.com/wolfi-dev/advisories/blob/main/trino.advisories.yaml#L100-L103

hectorj2f: Great! That is clear to me now!

PR #847, opened 2024-01-14
nack CVE-2023-41056 in redis-6.2

luhring:
A proactive defense in case the NVD CPEs end up representing the vulnerable range incorrectly (which happens often).


dlorenc (approved):

dlorenc:
Nice!

PR #852, opened 2024-01-15
Adding CVEs for prometheus-bitnami-compat

chainguardian:
Adding Advisory GHSA-xwh9-gc39-5298 for prometheus-bitnami-compat


cpanato:
added in https://github.com/wolfi-dev/advisories/pull/851

PR #958, opened 2024-01-24
Adding fixed CVEs for istio-install-cni-1.19

chainguardian:
Adding Fixed Advisory GHSA-7f9x-gw85-8grf for istio-install-cni-1.19


cpanato:
added here https://github.com/wolfi-dev/advisories/pull/955

PR #985, opened 2024-01-25
Create thanos-0.33.advisories.yaml

stgarf:
Adding Fixed Advisory https://github.com/advisories/GHSA-m425-mq94-257g and https://github.com/advisories/GHSA-qppj-fm5r-hxr3 for thanos-0.33

### DIFF
--- a/thanos-0.33.advisories.yaml
+++ b/thanos-0.33.advisories.yaml
@@ -1,0 +1,16 @@
+schema-version: 2.0.2
+
+package:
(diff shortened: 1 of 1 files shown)


stgarf:
Closing as not needed after a chat with @luhring 

`If that's a brand new version stream, and thus you're creating the first version of that particular package, you don't need to record fixes. For us, fixes mean we did ship a version of it that had the vuln, and a later version of the package is remediated`

PR #1026, opened 2024-01-29
pulumi-kubernetes-operator: add info about GHSA-449p-3h89-pw88

hectorj2f:
GHSA-449p-3h89-pw88 for pulumi-kubernetes-operator requires switching from `gopkg.in/src-d/go-git.v4` which has been archived now with `github.com/go-git/go-git/v5` that fixed the detected vulnerability.

### DIFF
--- a/pulumi-kubernetes-operator.advisories.yaml
+++ b/pulumi-kubernetes-operator.advisories.yaml
@@ -87,6 +87,13 @@ advisories:
             componentLocation: /usr/bin/pulumi-kubernetes-operator
             scanner: grype
 
+  - id: GHSA-449p-3h89-pw88
+    events:
+      - timestamp: 2024-01-29T16:01:39Z
+        type: pending-upstream-fix
+        data:
+          note: This vulnerability requires to replace gopkg.in/src-d/go-git.v4 by github.com/go-git/go-git/v5 which requires to modify the source code.
+
   - id: GHSA-6xv5-86q9-7xr8
     events:
       - timestamp: 2023-09-09T15:18:19Z


pulumi-kubernetes-operator.advisories.yaml
```diff
@@ -87,6 +87,13 @@ advisories:
             componentLocation: /usr/bin/pulumi-kubernetes-operator
             scanner: grype
 
+  - id: GHSA-449p-3h89-pw88
+    events:
+      - timestamp: 2024-01-29T16:01:39Z
+        type: pending-upstream-fix
+        data:
+          note: This vulnerability requires to replace gopkg.in/src-d/go-git.v4 by github.com/go-git/go-git/v5 which requires to modify the source code.
```
luhring, line 95: What was the nature of the code modification required? Is it something we could pursue and then leverage to get the the dep bumped to the fixed version?

hectorj2f: i don't understand your question. They are different modules.

luhring: > which requires to modify the source code

I'm trying to understand this part of the note — what does this mean exactly? Modify how?

hectorj2f: Correct me if I am wrong, but I assume you don't want to know which go files, function calls need to be updated when we are talking about modifying the code in pulumi-kubernetes-operator to use a different module that does similar stuff. Right ?

hectorj2f: I don't think so.. But anyone will need to update stack_controller.go file and in particular the function `revisionAtWorkingDir` where the code is using a function `PlainOpenWithOptions` of the archived/deprecated module `gopkg.in/src-d/go-git.v4`.

luhring: (In case it's helpful) I _think_ upstream might just need the new Go module swapped in w/ `go get ...`, and then something like this for the code:

```diff
--- a/pkg/controller/stack/stack_controller.go
+++ b/pkg/controller/stack/stack_controller.go
@@ -18,6 +18,7 @@ import (
        "sync"
        "time"

+       "github.com/go-git/go-git/v5"
        "github.com/operator-framework/operator-lib/handler"
        "github.com/pulumi/pulumi-kubernetes-operator/pkg/apis/pulumi/shared"
        pulumiv1 "github.com/pulumi/pulumi-kubernetes-operator/pkg/apis/pulumi/v1"
@@ -30,7 +31,6 @@ import (
        "github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
        "github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
        giturls "github.com/whilp/git-urls"
-       git "gopkg.in/src-d/go-git.v4"
        corev1 "k8s.io/api/core/v1"
        apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
        k8serrors "k8s.io/apimachinery/pkg/api/errors"
```

Ideally, we could submit this CVE fix upstream, and then carry a patch of that fix in Wolfi until the PR lands upstream. WDYT?

hectorj2f: I've opened this PR https://github.com/pulumi/pulumi-kubernetes-operator/pull/546

luhring: Awesome!

luhring: Friendly ping @hectorj2f — awesome job getting the fix merged upstream. Now we have to finish the work on our side. Let me know if you need help with that!

hectorj2f: Do we wait until they release the new version ? Or we change the note here specifying the pending new release with a fix.

luhring: Yeah good question, I think it's a judgment call. But I'd bias toward carrying the patch now until the next release, also considering that this item has been open for about a week already. So we'd update the package itself and then have a fixed advisory instead of `pending-upstream-fix`

hectorj2f: Closing this PR! I got the patch here: https://github.com/wolfi-dev/os/pull/12192

PR #1077, opened 2024-02-02
Adding advisories for apache spark

mamccorm:
Creating advisories for apache spark.
Note, as this is for a new package, will require the following to be merged first: https://github.com/wolfi-dev/os/pull/12309


spark.yaml
```diff
@@ -0,0 +1,157 @@
...
+
+  - id: CVE-2022-3510
+    aliases:
+      - GHSA-4gg5-vx3j-xwc7
+    events:
+      - timestamp: 2024-02-02T22:15:34Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-not-included-in-package
+          note: protobuf-java classes are not included in runtime jars, but its metadata is still listed in META-INF pom.xml and properties files
```
pdeslaur, line 120: I'm not sure I agree with this - I have found the protobuf classes in the shaded jar

luhring (changes requested):
(placing a hold until we sync next week!)

spark.yaml
mamccorm: Resolving as we paired up earlier and walked through some of the items, including this one :)

luhring:
Continuing this in #1112

PR #1117, opened 2024-02-06
lynx: add advisories

rawlingsj:


rawlingsj:
Opening as draft as I could do with a second set of eyes on these to confirm what I think. 

@luhring can you take a look when you get a few mins please?

luhring:
Thanks @rawlingsj! I'll look at this first thing when I get in today

lynx.advisories.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+  - id: CVE-2016-9179
+    aliases:
+      - GHSA-r2wx-qjhc-7h6p
+    events:
+      - timestamp: 2024-02-06T10:58:47Z
+        type: false-positive-determination
+        data:
+          type: vulnerability-record-analysis-contested
+          note: Contested see https://seclists.org/oss-sec/2016/q4/344
```
luhring, line 25: I could be wrong, but it looks the dissenting message there was deleted, and that the rest of the participants (including what looks like the maintainer) think this is a legitimate issue. FWIW I also tried the PoC URL on Go, and Go handles it correctly.

All that said, I'm guessing there's a decent chance this actually has been patched by now, and that the vuln match is only turning up because of the CPE data missing version ranges. We should confirm that.

lynx.advisories.yaml
```diff
@@ -0,0 +1,25 @@
...
+advisories:
+  - id: CVE-2012-5821
+    aliases:
+      - GHSA-px54-5fmq-rf2f
+    events:
+      - timestamp: 2024-02-06T10:45:42Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-not-in-execution-path
+          note: We build lynx against OpenSSL, not GnuTLS which is the affceted execution path
```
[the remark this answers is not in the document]
luhring: Otherwise this looks great to me

lynx.advisories.yaml
luhring: My read based on https://lists.debian.org/debian-lts/2016/11/msg00082.html is that this was fixed in version `2.8.9dev.11`. If you agree, let's update FP type to `vulnerable-code-version-not-used`, and update the note to explain which version fixed this (prior to Wolfi packaging), and maybe even link to that discussion thread in case anyone wants to verify our thought process.

cpanato:
is this still valid?

cpanato:
@rawlingsj should we close this?

PR #1131, opened 2024-02-08
Fix validation errors for new duplicate validation check

luhring:
This incorporates fixes for the new duplicate advisory ID check in wolfictl: https://github.com/wolfi-dev/wolfictl/pull/617

This also catches up on alias discovery (which isn't checked in CI currently).

There are some necessary validation failures due to the nature of this fix. Here's what's expected:

```console
❌ advisory data is not valid.

invalid change(s) in diff:
    cluster-autoscaler-1.25:
        GHSA-f9jg-8p32-2f55:
            advisory was removed
    gitlab-shell:
        GHSA-45x7-px36-x8w8:
            advisory was removed
        CVE-2023-48795:
            event 1 (just added):
                event's timestamp (2023-12-29T21:35:56Z) set to more than 3 days ago; timestamps should accurately capture event creation time
            event 2 (just added):
                event's timestamp (2023-12-29T21:36:27Z) set to more than 3 days ago; timestamps should accurately capture event creation time
```

### DIFF
--- a/management-api-for-apache-cassandra.advisories.yaml
+++ b/management-api-for-apache-cassandra.advisories.yaml
@@ -11,7 +11,7 @@ advisories:
       - timestamp: 2024-01-14T18:15:03Z
         type: pending-upstream-fix
         data:
-          note: "To fix the CVE, we have to upgrade 'swagger-jaxrs2' to '2.2.0' or later but this fix will require some code change since the upgrade cause to build fail due to compilation errors like: 'src/main/java/com/datastax/mgmtapi/resources/LifecycleResources.java:[425,28] cannot access com.fasterxml.jackson.core.exc.StreamWriteException'"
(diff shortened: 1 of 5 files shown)


management-api-for-apache-cassandra.advisories.yaml
```diff
@@ -11,7 +11,7 @@ advisories:
       - timestamp: 2024-01-14T18:15:03Z
         type: pending-upstream-fix
         data:
-          note: "To fix the CVE, we have to upgrade 'swagger-jaxrs2' to '2.2.0' or later but this fix will require some code change since the upgrade cause to build fail due to compilation errors like: 'src/main/java/com/datastax/mgmtapi/resources/LifecycleResources.java:[425,28] cannot access com.fasterxml.jackson.core.exc.StreamWriteException'"
+          note: 'To fix the CVE, we have to upgrade ''swagger-jaxrs2'' to ''2.2.0'' or later but this fix will require some code change since the upgrade cause to build fail due to compilation errors like: ''src/main/java/com/datastax/mgmtapi/resources/LifecycleResources.java:[425,28] cannot access com.fasterxml.jackson.core.exc.StreamWriteException'''
```
luhring, line 14: Yes, these look odd. But I'd prefer to keep these changes for now. I'd rather we have a consistent way of handling string quotes in the code throughout, and we can always change our approach in the future, while maintaining consistency at the same time.

pdeslaur (approved):

PR #1187, opened 2024-02-11
skaffold: pending-upstream-fix

hectorj2f:
We decided to mark these two vulnerabilities as `pending-upstream-fix` due to how behind upstream skaffold is at the moment.

Happy to add more details if you think it isn't enough.


cpanato (dismissed):

pdeslaur:
This should be pending-upstream-fix since skaffold is still supported!

pdeslaur (changes requested):
See comment above!

pdeslaur (approved):
🙏 🙏 🙏

PR #1709, opened 2024-03-01
opensearch-dashboard-2: remove fixed events as we have withdrawn 2.12.0

rawlingsj:
https://github.com/wolfi-dev/os/pull/13877


rawlingsj:
The validate check is expected to fail as we are indeed deleting events because we withdrew the fixed version from Wolfi

PR #3125, opened 2024-03-20
go: File advisories for EOL packages

pdeslaur:
These packages are no longer supported by the Go team, and there are no fixes available from upstream: https://groups.google.com/g/golang-announce/c/5pwGVUPoMbg


cpanato (approved):
Thx

PR #3899, opened 2024-04-16
Adding Advisory GHSA-2gvw-w6fj-7m3c for argo-cd-2.7

cpanato:


hectorj2f (approved):

hectorj2f (changes requested):
Could we add a fix-not-planned event too ? That way we don't have to open a separate PR for this one.

```
timestamp: 2024-03-30T16:47:59Z
        type: fix-not-planned
        data:
          note: Argo CD 2.7 is longer supported. Security Support ended on 2024-02-05.
          ```

PR #3910, opened 2024-04-16
Adv telegraf 1.26 1.26.3 r11

cpanato:


telegraf-1.26.advisories.yaml
```diff
@@ -75,6 +75,23 @@ advisories:
           type: vulnerable-code-not-included-in-package
           note: Only affects Windows
 
+  - id: CVE-2023-45288
```
hectorj2f, line 78: We should add the fix-not-planned events at the same time. Telegraf is no longer supported upstream.

PR #3930, opened 2024-04-18
kots advisory

sergio-chainguard:


sergio-chainguard:
There is already a comment about this CVE

PR #4376, opened 2024-04-27
Add advisory data for recently detected false positives

hectorj2f:


luhring:
This PR looks good — just a quick question: I see there are no detection events before these new FP events, is that expected? The detection data might be nice to have to give context to this false positive, but just curious here

hectorj2f:
@luhring We now have all the detections.

luhring (approved):
Woohoo! Thanks!

PR #4474, opened 2024-05-01
update advisory to pending-upstream-fix

mamccorm:
See:
https://github.com/chainguard-dev/image-release-stats/issues/108


hectorj2f (approved):

cpanato (commented):
dont we need to remove the fix entry?

@rawlingsj

mamccorm:
Hey @cpanato, the way it was explained to me in the past, we don't tamper with history and add new timestamp events for any changes. But if thats incorrect let me know

PR #4485, opened 2024-05-02
haproxy 2.9

developer-guy:
/cc @mamccorm


hectorj2f (approved):
We can close the PR. Let's rely on our automation service to create the detection events with all the details.

PR #4528, opened 2024-05-06
adds keycloak-operator advisory

mritunjaysharma394:
I think we already have a similar [CVE adv for keycloak](https://github.com/wolfi-dev/advisories/blob/ec6438d28576369664567a2bf5442691fb6d02b6/keycloak.advisories.yaml#L362), this one is also needed for the keycloak-operator as the upstream fix has still to be released


mritunjaysharma394:
Closing as it is already merged

PR #4570, opened 2024-05-07
Modified neuvector-controller advisory

Pyronewbic:


neuvector-controller-5.3.advisories.yaml
```diff
@@ -9,31 +9,16 @@ advisories:
       - GHSA-622h-h2p8-743x
     events:
       - timestamp: 2024-04-27T10:34:17Z
-        type: detection
```
rawlingsj, line 12: thanks - though we can keep these existing `detection` events, and append the `false-positive-determination` below it.

neuvector-controller-5.3.advisories.yaml
```diff
@@ -9,31 +9,16 @@ advisories:
...
+          type: vulnerable-code-version-not-used
+          note: neuvector-controller is using a non vulnerable version of neuvector module which has been released after the affected versions.
 
   - id: CVE-2024-3154
     aliases:
-      - GHSA-c5pj-mqfh-rvc3
+      - GHSA-2cgq-h8xw-2v5j
     events:
-      - timestamp: 2024-04-27T10:34:18Z
-        type: detection
```
rawlingsj, line 29: similar to above, we should keep the `detection` event and append a new `fixed` event after.  We don't need to remove previous events.

rawlingsj (changes requested):

rawlingsj:
not needed after https://github.com/wolfi-dev/os/pull/18873

PR #4623, opened 2024-05-08
Add CVE data for terraform, consul and k8ssandra-operator

hectorj2f:


hectorj2f:
@luhring I've verified the changes using git and scanning them with govulncheck, and removed calico from my initial commit as I didn't check it yet. Instead I added docker which I've verified. We are good to merge this PR in my opinion.

luhring (approved):
Thanks @hectorj2f!

PR #4923, opened 2024-05-15
chore: Add advisories for Firefox

EyeCantCU:


mamccorm:
Closing this for now, as it looks like the automation ran and created detection events - so we'll need to update the detection events separately

PR #4953, opened 2024-05-17
file advisories for werkzeug vulnerabilities in airflow

srishtih:


mritunjaysharma394 (commented):
I think we might have to force merge this one since this is a CI Failure that is a false alarm since I guess it is thinking connexion is a misspelling but it is not https://pypi.org/project/connexion/

PR #5147, opened 2024-06-03
octo-sts advisory update

ritster:
[GHSA-75r6-6jg8-pfcq](https://github.com/advisories/GHSA-75r6-6jg8-pfcq) is not an issue for octo-sts, as Wolfi uses v0.2.0 and the vulnerability is only present in v<0.1.0.


pdeslaur (approved):
Nice!

PR #5185, opened 2024-06-05
minio advisory update (false positive)

ritster:


minio.advisories.yaml
```diff
@@ -286,6 +286,11 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2024-06-05T14:30:56Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-version-not-used
+          note: Vulnerable version of minio is <0.20240527.191746 and wolfi uses 0.20240528.171904.
```
pdeslaur, line 293: We shoud file a "fixed" advisory in this case since this advisory affects an APK package (`componentType: apk`)

ritster: Can do. Why is that the case?

hectorj2f (approved):

PR #5201, opened 2024-06-06
fix(openssl): CVE-2024-4603 was fixed not nacked

luhring:
cc: @cpanato @hectorj2f 

Expected validation result:

```text
invalid change(s) in diff:
    openssl:
        CGA-9c28-ww88-6g95:
            one or more events were removed
```

### DIFF
--- a/openssl.advisories.yaml
+++ b/openssl.advisories.yaml
@@ -121,23 +121,6 @@ advisories:
         type: fixed
         data:
           fixed-version: 3.3.0-r8
-      - timestamp: 2024-05-30T07:20:32Z
-        type: detection
-        data:
-          type: scan/v1
-          data:
-            subpackageName: openssl
-            componentID: 3a54114a51f84e92
-            componentName: openssl
-            componentVersion: 3.3.0
-            componentType: binary
-            componentLocation: /usr/bin/openssl
-            scanner: grype
-      - timestamp: 2024-05-31T10:19:42Z
-        type: false-positive-determination
-        data:
-          type: vulnerable-code-not-included-in-package
-          note: This code has been manually patched using the upstream patches from commit (85ccbab216da245cf9a6503dd327072f21950d9b).
 
   - id: CGA-c2c3-7c62-3x6m
     aliases:


cpanato (approved):

cpanato (commented):
i am thinking, the automation might open the detection again

because when running the scan i got

```
$ wolfictl scan -r openssl
📡 Finding remote packages
🔎 Scanning "/var/folders/kl/q9mydw095ln5s7wj971qcrx40000gn/T/x86_64-openssl-3.3.0-r9-3269331789.apk"
└── 📄 /usr/bin/openssl
        📦 openssl 3.3.0 (binary)
            Unknown CVE-2024-4603

🔎 Scanning "/var/folders/kl/q9mydw095ln5s7wj971qcrx40000gn/T/aarch64-openssl-3.3.0-r9-3052073876.apk"
└── 📄 /usr/bin/openssl
        📦 openssl 3.3.0 (binary)
            Unknown CVE-2024-4603
```

hectorj2f:
As mentioned on our discussion, we might need to stick to false positives otherwise we will get a detection event again.

luhring:
> As mentioned on our discussion, we might need to stick to false positives otherwise we will get a detection event again.

The scanner match shown above is for a "binary" type package. This **would** cause a detection event to get created, but the better solution here is to remove "binary" packages from consideration in our vulnerability matching, which is more inline with how Grype behaves when scanning images. That's addressed in https://github.com/wolfi-dev/wolfictl/pull/945.

Once we regain parity with Grype with the above PR, switching from "false positive" to "fixed" should work great, because it will compare the APK (and not any binary packages) to our data, and openssl versions at or above `3.3.0-r8` will correctly show as fixed, and versions less than that will correctly show as not fixed. This latter piece won't be correct for as long as the "false positive" status remains in place.

hectorj2f:
> Once we regain parity with Grype with the above PR, switching from "false positive" to "fixed" should work great,

Sounds good to me.

hectorj2f (approved):

PR #5640, opened 2024-06-10
Updated advisory data for Percona

kz3640:


luhring:
Any chance we know specifically which version or git commit fixed this, that we could include in the note? Also curious if we know why the scanner found this!

kz3640:
I'm not able to find exact documentation for CVE-2021-27928 regarding Percona for mySQL. The closest I could find was this documentation about it being fixed in Percona for mySQL with XtraDB here https://docs.percona.com/percona-xtradb-cluster/8.0/release-notes/Percona-XtraDB-Cluster-8.0.22-13.1.html

luhring (approved):

PR #5846, opened 2024-06-16
kyverno: add adv for square/go-jose

Dentrax:


kyverno.advisories.yaml
```diff
@@ -309,6 +309,10 @@ advisories:
         type: fixed
         data:
           fixed-version: 1.11.4-r5
+      - timestamp: 2024-06-16T21:29:25Z
+        type: fix-not-planned
```
hectorj2f, line 313: I believe this is more of a pending-upstream-fix rather than fix-not-planned here.

Dentrax: Thanks. Done.

hectorj2f (approved):

PR #6072, opened 2024-06-25
Adds a false positive event for mariadb-test CVE-2024-1597

dakaneye:
https://github.com/wolfi-dev/os/actions/runs/9651623275/job/26621058665?pr=22539

Found the CVE reported in the above OS PR check and did some digging: 
* In the upstream repository, they are pulling in a hardcoded postgres jar file for a test (really old version) that _shouldn't_ be in the live execution path for mariadb
* We _could_ even consider removing the test package as there are no public OS packages depending on it.

### DIFF
--- a/mariadb-10.11.advisories.yaml
+++ b/mariadb-10.11.advisories.yaml
@@ -4,12 +4,12 @@ package:
   name: mariadb-10.11
 
 advisories:
-  - id: CGA-69ff-4xff-qv2w
+  - id: CGA-4p65-rcvr-255v
     aliases:
-      - CVE-2020-13692
-      - GHSA-88cc-g835-76rp
+      - CVE-2022-21724
+      - GHSA-v7wg-cpwc-24m4
     events:
-      - timestamp: 2024-01-24T16:39:12Z
+      - timestamp: 2024-01-24T16:39:13Z
         type: detection
         data:
           type: scan/v1
@@ -22,12 +22,12 @@ advisories:
             componentLocation: /usr/mysql-test/plugin/connect/connect/std_data/JdbcMariaDB.jar
             scanner: grype
 
-  - id: CGA-4p65-rcvr-255v
+  - id: CGA-69ff-4xff-qv2w
     aliases:
-      - CVE-2022-21724
-      - GHSA-v7wg-cpwc-24m4
+      - CVE-2020-13692
+      - GHSA-88cc-g835-76rp
     events:
-      - timestamp: 2024-01-24T16:39:13Z
+      - timestamp: 2024-01-24T16:39:12Z
         type: detection
         data:
           type: scan/v1
@@ -76,3 +76,8 @@ advisories:
             componentType: java-archive
             componentLocation: /usr/mysql-test/plugin/connect/connect/std_data/JdbcMariaDB.jar
             scanner: grype
+      - timestamp: 2024-06-25T18:11:51Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-not-in-execution-path
+          note: This CVE appears to stem from a hardcoded postgres jar file used specifically for a stale test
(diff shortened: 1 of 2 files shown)


mariadb-10.11.advisories.yaml
```diff
@@ -4,12 +4,12 @@ package:
   name: mariadb-10.11
 
 advisories:
-  - id: CGA-69ff-4xff-qv2w
```
dakaneye, line 7: I'm not sure why these got updated...

smoser:
I'm interested in others thoughts on this.

@luhring , I don't like declaring "this is not in your execution path".   Reality is that no one is going to consume the '-test' package that this file is present, but 
1. if we accidentally moved it to the main package then it would find its way into an image (and it would be ignored?)
2. if someone _does_ consume -test package, then the it is probably _not_ a false positive.

Maybe there are other examples of this?

hectorj2f (approved):

hectorj2f:
@smoser It looks like this code is mostly intended to be used during development [here](https://github.com/MariaDB/server/blob/db06c5dd07e6ffd6a569afe6950fa77c2f02f811/storage/connect/jdbconn.cpp#L759). It also looks awkward to use this version of postgresql released on `Feb 19, 2016`. I feel this is not used at all.

hectorj2f:
I am also curious to know how grype detects the usage of this postgresql version as it is not set or defined anywhere else.

hectorj2f:
I saw this reference to a newer version of postgresql used in a khttps://github.com/MariaDB/server/blob/7c5fdc9b6a45f4e8bad1ae5c7c93b457be9891fa/storage/connect/mysql-test/connect/t/jdbc_postgresql.test#L6. This makes more sense to me rather than this reported old version detected by the scanner.

mariadb-10.11.advisories.yaml
luhring: It's just a reordering. It looks weird. This is a known vestigial thing that affects some advisories files. There's a command to see what's actually being changed, e.g. I just ran it for your branch:

```console
$ wolfictl adv diff
Auto-detected distro: Wolfi

( - removed / ~ modified / + added )

~ document "mariadb-10.11"
  ~ advisory "CGA-x53v-hg6v-7rvv"
    + event "false-positive-determination" @ 2024-06-25T18:11:51Z
~ document "mariadb-11.2"
  ~ advisory "CGA-5cmg-w6mr-wx73"
    + event "false-positive-determination" @ 2024-06-25T18:13:23Z
```

smoser:
@hectorj2f , I absolutely agree that the code is not in an execution path, butI think we came to the conclusion that we should just drop the mariadb-test package which was delivering this.  

my concern  below is not actually valid as the path would change to not be /usr/mysql-test/plugin/connect/connect/std_data/JdbcMariaDB.jar if it moved to main.

> 1. if we accidentally moved it to the main package then it would find its way into an image (and it would be ignored?)

mariadb-10.11.advisories.yaml
```diff
@@ -76,3 +76,8 @@ advisories:
             componentType: java-archive
             componentLocation: /usr/mysql-test/plugin/connect/connect/std_data/JdbcMariaDB.jar
             scanner: grype
+      - timestamp: 2024-06-25T18:11:51Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-not-in-execution-path
+          note: This CVE appears to stem from a hardcoded postgres jar file used specifically for a stale test
```
luhring, line 83: I'd like to get confirmation that this is still the case with the recent related package PR — and if not we can update this type and note

mariadb-11.2.advisories.yaml
```diff
@@ -60,6 +60,11 @@ advisories:
             componentType: java-archive
             componentLocation: /usr/mariadb-test/plugin/connect/connect/std_data/JdbcMariaDB.jar
             scanner: grype
+      - timestamp: 2024-06-25T18:13:23Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-not-in-execution-path
+          note: This CVE appears to stem from a hardcoded postgres jar file used specifically for a stale test
```
luhring, line 67: ditto the above!

luhring (changes requested):

dakaneye:
Looks like we addressed by removing the mariadb-test subpackage entirely, and we'll let automation handle this later.

PR #6130, opened 2024-06-27
chore: remove cluster-autoscaler advisory files

hectorj2f:
These packages have been moved to enterprise-packages.


cpanato:
need rebase

hectorj2f:
Done

cpanato:
Were the packages withdrawn as well? otherwise, i think we keep those here

thoughts @luhring ?

luhring:
> Were the packages withdrawn as well? otherwise, i think we keep those here
> 
> thoughts @luhring ?

I think this kind of cleaning up is the right move. If the package is no longer maintained in Wolfi, and specifically if it's been moved to another package repo like enterprise, let's make sure the enterprise advisory data has what it needs and then remove it from Wolfi advisory data.

This helps us move toward a "clean" state where there's no duplication of a package's advisory data across the repos. And soon I'm hoping we'll consolidate to a single git repo.

Thanks @hectorj2f!

hectorj2f:
Everything is clean and ready for removal of these advisory files.

luhring (approved):

cpanato:
good to know

PR #6138, opened 2024-06-27
mattermost-9: update advisory

Dentrax:


hectorj2f (approved):

hectorj2f:
@Dentrax It seems we removed some CVEs.

hectorj2f (changes requested):
Restore all the removed CVEs.

luhring:
@Dentrax I'm wondering if you tried to overwrite the existing file here in Wolfi advisories altogether. If so, know that _we can't do that_, since that's going to delete advisories we've already started.

It sounds like what we mean to do here is to just add the conclusion events (like false positives, and pending upstream fix) from the extra advisories to this file. That will have to be done by hand for now (this seems like a one-off accident scenario, based on the context I'm seeing in Slack).

Once that's done, I'm expecting we'll remove the package from extra packages and the advisories file from extra advisories. Let me know if I'm mistaken here!!

hectorj2f:
It seems there aren't any changes.

hectorj2f:
@Dentrax This LGTM, I am gonna use this PR for some testing if you don't mind. I will take care of it.

hectorj2f (approved):

PR #6585, opened 2024-07-06
kyverno-1.12: version stream

Dentrax:


ajayk (changes requested):
Package name should reflect the version stream

Dentrax:
Not needed since automation handled this already! https://github.com/wolfi-dev/advisories/blob/main/kyverno-1.12.advisories.yaml

PR #7310, opened 2024-08-23
php-8.2: file advisories for php-calendar

imjasonh:


php-8.2.advisories.yaml
```diff
@@ -94,6 +94,15 @@ advisories:
...
           note: This CVE targets a PHP-based web application called "PHP Calendar," and is unrelated to the PHP calendar extension.
+      - timestamp: 2024-08-23T17:36:00Z
+        type: false-positive-determination
+        data:
+          type: component-vulnerability-mismatch
+          note: The vulnerability affects https://github.com/sproctor/php-calendar which is a different project from the official PHP calendar extension.
+      - timestamp: 2024-08-23T17:36:06Z
+        type: fix-not-planned
+        data:
+          note: '"Taking that the impact of this issue is low-to-none due to very specific pre-conditions - the decision was not to fix it at this time." - https://bugzilla.redhat.com/show_bug.cgi?id=12287'
```
imjasonh, line 105: I'm open to improving this note, I also found a recent PHP upstream issue that notes this is a perpetual match according to Grype, and indicates there is no plan to fix this: https://github.com/php/php-src/issues/14050

luhring: Were these two events meant to be for separate advisories?

imjasonh: Yes!

imjasonh:
These are already filed at head.

PR #7341, opened 2024-08-30
Add advisories for py3.11 for the -dev variant of the node image

mamccorm:
Our -dev image variant contains python 3.11. We're reviewing this in future and may remove this, but will require some notifications. This PR copies over the advisories already raised for CVE-2024-7592 and CVE-2024-6923 for python 3.11, and associates them with the node packages (that are still active and supported).


philroche (approved):

cpanato (approved):
lgtm

PR #7448, opened 2024-09-09
drop loki-3

imjasonh:
https://github.com/wolfi-dev/os/pull/27968


cpanato:
@imjasonh did you withdraw the packages? if not we will keep this for the ones that use the old package

imjasonh:
> @imjasonh did you withdraw the packages? if not we will keep this for the ones that use the old package

Yeah I think that's the way I'm leaning too. I'll just close this, and if loki-3.1 adds any more advisories we can add those in a new file.

PR #8228, opened 2024-09-14
podman: GHSA-874v-pj72-92f3

hectorj2f:
This is a clear case of false positive given that the detected version is older than the currently supported by the package.

### DIFF
--- a/podman.advisories.yaml
+++ b/podman.advisories.yaml
@@ -1,0 +1,16 @@
+schema-version: 2.0.2
+
+package:
+  name: podman
+
+advisories:
+  - id: CGA-pg8c-46w3-4rmx
+    aliases:
+      - CVE-2024-1753
+      - GHSA-874v-pj72-92f3
+    events:
+      - timestamp: 2024-09-14T19:09:03Z
+        type: false-positive-determination
+        data:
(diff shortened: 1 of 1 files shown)


podman.advisories.yaml
```diff
@@ -0,0 +1,16 @@
...
+advisories:
+  - id: CGA-pg8c-46w3-4rmx
+    aliases:
+      - CVE-2024-1753
+      - GHSA-874v-pj72-92f3
+    events:
+      - timestamp: 2024-09-14T19:09:03Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-not-included-in-package
```
luhring, line 15:
```suggestion
          type: vulnerable-code-version-not-used
```

luhring (dismissed):

luhring:
Hi @hectorj2f! — I think I'm confused by this series of events on this PR: I made a suggestion to fix the false positive classification, you accepted and committed the suggestion, I approved, then you reverted the classification back to the original state. Was that an accident, or are you suggesting that's a better FP type here? Thanks! 🙏

hectorj2f:
That's due to a bad rebase. My mistake. I'll fix it.

luhring (approved):

PR #8233, opened 2024-09-15
Cleanup renamed packages

hectorj2f:
We need to define procedures on which we remove this files or moved them to a different repository.


luhring:
Overriding to fix validation errors

luhring (approved):

PR #8240, opened 2024-09-16
GHSA-h4gh-qq45-vh27 superset advisory update

jamie-albert:
Superset: Upstream has pinned version of python package cryptography [as seen here](https://github.com/apache/superset/blob/f0971a850c12c69b5f09874ab50b25976788a6de/requirements/base.txt#L83). Changes in 42.x.x to 43.x.x require upstream changes.'


philroche:
Isn't this an opportunity to patch the `requirements/base.txt` file?

jamie-albert:
The pinned version was changed to what it currently is within the last two weeks, when two versions of cryptography have been released prior to that. This leads me to believe they are deliberately holding the version back.

philroche (approved):

PR #8242, opened 2024-09-16
airflow GHSA-fw5r-6m3x-rh7p advisory update

jamie-albert:
Due to the tightly coupled nature of airflow and Flask-AppBuilder [stated here](https://github.com/apache/airflow/blob/30925c739b60d8a54d84c7c58a3ab854c167f2c1/airflow/providers/fab/provider.yaml#L50), any changes to the security/manager.py file need to be implemented in an [override file found here](https://github.com/apache/airflow/blob/main/airflow/providers/fab/auth_manager/security_manager/override.py). a [PR has been opened to suggest the changes](https://github.com/apache/airflow/pull/42261) however that is waiting upstream approval


mamccorm (approved):
Nice write-up and contribution upstream

PR #8367, opened 2024-09-19
add advisory entries for packages

hectorj2f:


rustup.advisories.yaml
```diff
@@ -20,6 +20,10 @@ advisories:
             componentType: rust-crate
             componentLocation: /usr/bin/rustup-init
             scanner: grype
+      - timestamp: 2024-09-19T08:52:38Z
+        type: pending-upstream-fix
+        data:
+          note: This vulnerability requires upstream changes. The maintainers pinned the openssl version due to build issues.
```
philroche, line 26: Can we link to where they tried this?

philroche (changes requested):

rustup.advisories.yaml
hectorj2f: @philroche I'd be happy to add the link but I don't think this should be a block comment.

philroche (approved):

rustup.advisories.yaml
philroche: approved but I do think information like that is needed for users to have trust in our determinations/decisions.

PR #8497, opened 2024-09-30
mattermost: GHSA-59hf-mpf8-pqjh is a false positive

hectorj2f:
There is a version mismatch in the detection of this CVE caused by grype.


philroche (approved):
Lint is failing due to whitespace. +1 pending resolving lint failures

PR #8501, opened 2024-09-30
confluent-kafka GHSA-735f-pc8j-v9w8 advisory

jamie-albert:
Due to the nature of protobuf being a transitive dependency in the confluent-kafka project, which can be seen as the only reference to protobuf in the build.gradle, versions.gradle or gradle.properties files is in the build.gradle, moving the [protobuf package within the shared JAR](https://github.com/confluentinc/kafka/blob/03095817ba4083115063a1df964d3a290406d167/build.gradle#L1855). We must wait for the dependency to be bumped upstream.

### DIFF
--- a/confluent-kafka.advisories.yaml
+++ b/confluent-kafka.advisories.yaml
@@ -21,6 +21,10 @@ advisories:
             componentType: java-archive
             componentLocation: /usr/lib/kafka/libs/protobuf-java-3.23.4.jar
             scanner: grype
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
Thanks https://github.com/wolfi-dev/advisories/pull/8488 is tracking the same for kafka but I like this description better for sure. :)

jamie-albert:
Effefctively the same thing has been approved by @srishtih in the above PR linked by @kranurag7

PR #8563, opened 2024-10-05
GHSA-2326-pfpj-vx3h - parsable: pending-upstream-fix advisory

mamccorm:
Filing a ending-upstream-fix advisory for GHSA-2326-pfpj-vx3h, which relates to the parsable package, and one of it's dependencies: lexical-core.

----------

**After this is approved / merged**, please close the following PR and delete the associated branch:
 - https://github.com/wolfi-dev/os/pull/29873

### DIFF
--- a/parseable.advisories.yaml
+++ b/parseable.advisories.yaml
@@ -42,6 +42,14 @@ advisories:
             componentType: rust-crate
             componentLocation: /usr/bin/parseable
             scanner: grype
+      - timestamp: 2024-10-05T21:41:06Z
+        type: pending-upstream-fix
+        data:
+          note: |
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
there was a release cut of 53.1.0 via https://github.com/apache/arrow-rs/pull/6501 

the main workspace version is also 53.1.0 now (https://github.com/apache/arrow-rs/blob/master/Cargo.toml) but I think this has not been updated to crates.io till now which still shows 53.0.0 (https://crates.io/crates/arrow-json) 

approving meanwhile given crates.io is not updated for the crate that relates to the GHSA.

PR #8569, opened 2024-10-07
adv(false positive): all the CVE here is marked false positive by main…

debasishbsws:
…tainers
https://github.com/apache/cassandra/blob/cassandra-5.0/.build/owasp/dependency-check-suppressions.xml

The way they suppressing false positive explained here https://jeremylong.github.io/DependencyCheck/general/suppression.html


debasishbsws:
@jamie-albert did you look into these?

jamie-albert (approved):

PR #8646, opened 2024-10-11
druid/GHSA-r7pg-v2c8-mfg3 advisory update

jamie-albert:
The remaining avro dependency that gets flagged with this CVE is a transitive dependency that is part of the hadoop-client-runtime.jar and requires a fix from the upstream maintainers. This advisory relates to the remediation found in the [following PR](https://github.com/wolfi-dev/os/pull/30615)


debasishbsws:
may require timestamp update

PR #8690, opened 2024-10-16
druid/GHSA-78wr-2p64-hpwj advisory udpate

jamie-albert:
The remaining commons-io dependencies that exist in the druid package are brought as transitive dependencies. For commons-io v2.4.0 this is ambari-metrics-common-2.7.0.0.0.jar and commons-io v2.8.0 is brought in via velocity-engine-core-2.3.jar, hadoop-client-runtime-3.3.4.jar / v3.3.6. These dependencies are not able to be upgraded to a higher version and require upstream maintainers to implement.


powersj (approved):
+1 as watching this work and investigation live

PR #8759, opened 2024-10-25
feat(sustaining): libreoffice-24.8 advisory - CVE-2012-5639 false-positive

ritster:


ritster:
Migrated this advisory over to the libreoffice-24.8 package from other libreoffice packages: https://github.com/wolfi-dev/advisories/blob/main/libreoffice-24.advisories.yaml, https://github.com/wolfi-dev/advisories/blob/main/libreoffice-24.2.advisories.yaml.

PR #8790, opened 2024-10-29
Eol CVE triage policy

hectorj2f:
Eol CVE triage policy for production


.github/chainguard/lifecycle-eol-cve-triage.sts.yaml
```diff
@@ -1,7 +1,8 @@
 issuer: https://accounts.google.com
 
 # prod-images: bot-triageolwolfi@staging-enforce-cd1e.iam.gserviceaccount.com
-subject: "108215247871814150708"
+# bot-triageolwolfi@prod-enforce-fabc.iam.gserviceaccount.com
+subject: "(102425108416137555121|108215247871814150708)"
```
ajayk, line 5: does subject work for multiple  items 
or it should be subject_pattern ?

hectorj2f: good call! I have to use subject_pattern here 👍🏻

ajayk (approved):

PR #8820, opened 2024-11-03
delete gatekeeper-3.16

hectorj2f:
We already have an advisory file on the enterprise-advisories repository.


hectorj2f:
We are keeping the advisory files for the moment due to keep some historical data.

PR #9033, opened 2024-11-10
Submitting pending-upstream-fix advisory for nutshell: GHSA-2326-pfpj…

mamccorm:
Attempting to upgrade this results in build issues. There are other dependencies which expect an older version of lexical-core. Raising as pending-upstream-fix

### DIFF
--- a/nushell.advisories.yaml
+++ b/nushell.advisories.yaml
@@ -20,3 +20,10 @@ advisories:
             componentType: rust-crate
             componentLocation: /usr/bin/nu_plugin_polars
             scanner: grype
+      - timestamp: 2024-11-10T22:09:11Z
+        type: pending-upstream-fix
+        data:
+          note: |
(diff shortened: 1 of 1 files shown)


nushell.advisories.yaml
```diff
@@ -20,3 +20,10 @@ advisories:
             componentType: rust-crate
             componentLocation: /usr/bin/nu_plugin_polars
             scanner: grype
+      - timestamp: 2024-11-10T22:09:11Z
+        type: pending-upstream-fix
+        data:
+          note: |
+            This vulnerability relates to the 'lexical-core' dependency, and is fixed in v1.0.0.
+            Unfortunately, nutshell has other dependencies which depend on an older version of lexical-core.
```
kranurag7, line 28: nit: we can update it to nushell. 
```suggestion
            Unfortunately, nushell has other dependencies which depend on an older version of lexical-core.
```

kranurag7 (approved):

philroche (approved):

PR #9137, opened 2024-11-18
Added EOL advisory for grafana-11.0/GHSA-29wx-vh33-7x7r

hbh7:
Following the same path as https://github.com/chainguard-dev/enterprise-advisories/pull/6724


hectorj2f (changes requested):
This is still an active version https://github.com/chainguard-dev/package-version-metadata/blob/main/version_streams/grafana.yaml#L117

PR #9196, opened 2024-11-21
lua-5.3/CVE-2020-15945 advisory update

jamie-albert:
## 1. **CVE-2020-1594**
- **pending-upstream-fix:** 
- **Details:** CVE affects versions lower than 5.4, consider upgrading to a higher version. There is no evidence of plan to not fix however there is no backport at this time.

### DIFF
--- a/lua5.3.advisories.yaml
+++ b/lua5.3.advisories.yaml
@@ -22,6 +22,6 @@ advisories:
             componentLocation: /.PKGINFO
             scanner: grype
       - timestamp: 2024-11-21T16:15:31Z
-        type: pending-upstream-fix
+        type: fix-not-planned
         data:
           note: CVE affects versions lower than 5.4, Lua will not ship any fixes for Lua-5.3.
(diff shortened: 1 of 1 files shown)


luhring:
Is it possible Lua 5.3 isn't supported any more and shouldn't still exist here?

I saw at https://www.lua.org/versions.html that they say:

>There will be no further releases of Lua 5.3.

To me that implies there's no upstream fix to wait for (re: `pending-upstream-fix` event type), and that perhaps there's a larger issue here with this package.

Curious for others' takes!

jamie-albert:
@luhring nice catch, would annotate fix-not-planned instead and investigate [why it is still is being used here in apk-tools?](https://github.com/wolfi-dev/os/blob/334e409225541cf5f4c6dc36bc5c0890cc55aee0/apk-tools.yaml#L38) we have lua-5.4 internally so I do not understand that.

luhring (approved):
Beautiful

PR #9300, opened 2024-12-04
mattermost-10.2 GHSA-q7pp-wcgr-pffx advisory update

jamie-albert:
## 1. **GHSA-q7pp-wcgr-pffx**
- **pending-upstream-fix:** 
- **Details:** The issue regarding disintegration/imaging v1.6.2 where the index of the scan function in scanner.go can go out of bounds has an open PR https://github.com/disintegration/imaging/issues/165 but no implanted fix yet

### DIFF
--- a/mattermost-10.2.advisories.yaml
+++ b/mattermost-10.2.advisories.yaml
@@ -867,6 +867,10 @@ advisories:
             componentType: go-module
             componentLocation: /usr/bin/mattermost
             scanner: grype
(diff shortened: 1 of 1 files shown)


debasishbsws:
I think the https://github.com/disintegration/imaging is not getting maintained anymore [last commit](https://github.com/disintegration/imaging/commits/master/) was 4 years ago. I think it would be a better idea if we can file an issue about this CVE if that is not already present.

debasishbsws (approved):

PR #9301, opened 2024-12-04
mattermost-10.2 GHSA-v42f-hq78-8c5m advisory update

jamie-albert:
## 1. **GHSA-v42f-hq78-8c5m**
- **false-positive-determination, component-vulnerability-mismatch:** 
- **Details:** This vulnerability [relates to versions of mattermost prior to v7.4.0](https://github.com/advisories/GHSA-v42f-hq78-8c5m), which is several releases old. Several sources tie the credit of the [individual contributing to the remediation](https://www.cve.org/CVERecord?id=CVE-2022-4045), the custom mattermost CVE ID (search for MMSA-2022-00124 [here](https://mattermost.com/security-updates/)) and the CVE ID found in the National Vulnerability Database which leads to the conculsion this is caused by an overly broad CPE configuration on the NVD record causing all versions of mattermost to become flagged.

### DIFF
--- a/mattermost-10.2.advisories.yaml
+++ b/mattermost-10.2.advisories.yaml
@@ -579,6 +579,11 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2024-12-04T14:41:29Z
+        type: false-positive-determination
+        data:
+          type: component-vulnerability-mismatch
+          note: 'This vulnerability relates to v7.x of mattermost, which is several releases old. The componentVersion is also being flagged incorrectly here by some scanners. A bug has been filed upstream against Syft, and the maintainers have confirmed it''s a scanner issue. See: https://github.com/anchore/syft/issues/2980.'
 
   - id: CGA-hxh6-329f-wm9m
     aliases:
--- a/mattermost-10.2.advisories.yaml
+++ b/mattermost-10.2.advisories.yaml
@@ -580,9 +580,9 @@ advisories:
             componentLocation: /.PKGINFO
             scanner: grype
       - timestamp: 2024-12-04T14:41:29Z
-        type: vulnerable-code-version-not-used
-        data:
-          type: component-vulnerability-mismatch
+        type: false-positive-determination
+        data: 
+          type: vulnerable-code-version-not-used
           note: 'This vulnerability relates to versions of mattermost prior to v7.4.0, which is several releases old. Several sources tie the credit of the individual contributing to the remediation, the custom mattermost CVE ID (MMSA-2022-00124) and the CVE ID found in the National Vulnerability Database which leads to the conculsion this is caused by an overly broad CPE configuration on the NVD record causing all versions of mattermost to become flagged. '
 
   - id: CGA-hxh6-329f-wm9m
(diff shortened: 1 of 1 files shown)


mattermost-10.2.advisories.yaml
```diff
@@ -579,6 +579,11 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2024-12-04T14:41:29Z
+        type: false-positive-determination
+        data:
+          type: component-vulnerability-mismatch
```
luhring, line 585: This isn't a `component-vulnerability-mismatch` since the vulnerability describes `mattermost`, which is the same component that the scanner found in the artifact.

`vulnerable-code-version-not-used` would be a more precise fit here.

mattermost-10.2.advisories.yaml
```diff
@@ -579,6 +579,11 @@ advisories:
...
+          note: 'This vulnerability relates to v7.x of mattermost, which is several releases old. The componentVersion is also being flagged incorrectly here by some scanners. A bug has been filed upstream against Syft, and the maintainers have confirmed it''s a scanner issue. See: https://github.com/anchore/syft/issues/2980.'
```
luhring, line 586: The component version flagged by Syft is `10.2.0-r0`, right? This match is to the APK (as seen in the detection event above), not a Go module — so I don't think that Syft issue is relevant here.

This match seems to be from an overly broad CPE configuration on the NVD record at https://nvd.nist.gov/vuln/detail/CVE-2022-4045, where where are no version constraints, so all versions of Mattermost are getting matched.

luhring (changes requested):

mattermost-10.2.advisories.yaml
```diff
@@ -579,6 +579,11 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2024-12-09T14:41:29Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-version-not-used
+          note: 'This vulnerability relates to versions of mattermost prior to v7.4.0, which is several releases old. Several sources tie the credit of the individual contributing to the remediation, the custom mattermost CVE ID (MMSA-2022-00124) and the CVE ID found in the National Vulnerability Database which leads to the conclusion this is caused by an overly broad CPE configuration on the NVD record causing all versions of mattermost to become flagged. '
```
luhring, line 586: I don't mean to be nit-picky, but I'm not sure what this sentence is saying exactly:

>Several sources tie the credit of the individual contributing to the remediation, the custom mattermost CVE ID (MMSA-2022-00124) and the CVE ID found in the National Vulnerability Database which leads to the conclusion this is caused by an overly broad CPE configuration on the NVD record causing all versions of mattermost to become flagged.

The part about broad CPE configuration is right, but I'm not sure how the phrases before that relate to that. I think we could just say that the CPE configuration has no version constraints, so it matches all version of mattermost, even though the vulnerability was fixed back in the 7.x versions.

jamie-albert: I think I made everything murky by trying to explain what I wrote in detail here. But it has no place in an advisory so I removed and simplified.

luhring (approved):

PR #9372, opened 2024-12-10
mattermost-10.2 advisory udpates

jamie-albert:
## 1. **GHSA-hqqj-g6mv-rw46**
- **false-positive-determination, vulnerable-code-version-not-used:** 
- **Details:** This vulnerability was remediated in mattermost v7.x, Specifically, in versions 7.1.4, 7.2.1, 7.3.1, and 7.4.0. For more information, please refer to: https://www.clouddefense.ai/cve/2022/CVE-2022-4019

## 2. **GHSA-rhh4-rh7c-7r5v**
- **pending-upstream-fix** 
- **Details:** This vulnerability relates to one of mattermost's dependencies, ‘github.com/mholt/archiver/v3'. Mattermost is running the most recent release of this dependency - v3.5, which still contains this vulnerability.

### DIFF
--- a/mattermost-10.2.advisories.yaml
+++ b/mattermost-10.2.advisories.yaml
@@ -538,7 +538,7 @@ advisories:
         type: false-positive-determination
         data:
           type: vulnerable-code-version-not-used
-          note: 'This vulnerability was remediated in mattermost v7.x, Specifically, in versions 7.1.4, 7.2.1, 7.3.1, and 7.4.0. For more information, please refer to: https://www.clouddefense.ai/cve/2022/CVE-2022-4019'
+          note: 'This vulnerability was remediated in mattermost v7.x, Specifically, in versions 7.1.4, 7.2.1, 7.3.1, and 7.4.0. For more information, please refer to https://mattermost.com/security-updates/ and search for the ID MMSA-2022-00118 in the server tab. ' 
 
   - id: CGA-hq9q-8hrj-882q
     aliases:
--- a/mattermost-10.2.advisories.yaml
+++ b/mattermost-10.2.advisories.yaml
@@ -538,7 +538,7 @@ advisories:
         type: false-positive-determination
         data:
           type: vulnerable-code-version-not-used
-          note: 'This vulnerability was remediated in mattermost v7.x, Specifically, in versions 7.1.4, 7.2.1, 7.3.1, and 7.4.0. For more information, please refer to https://mattermost.com/security-updates/ and search for the ID MMSA-2022-00118 in the server tab. ' 
+          note: 'This vulnerability was remediated in mattermost v7.x, Specifically, in versions 7.1.4, 7.2.1, 7.3.1, and 7.4.0. For more information, please refer to https://mattermost.com/security-updates/ and search for the ID MMSA-2022-00118 in the server tab. '
 
   - id: CGA-hq9q-8hrj-882q
     aliases:
--- a/mattermost-10.2.advisories.yaml
+++ b/mattermost-10.2.advisories.yaml
@@ -534,6 +534,11 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2024-12-10T05:24:48Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-version-not-used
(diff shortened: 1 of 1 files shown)


mattermost-10.2.advisories.yaml
```diff
@@ -534,6 +534,11 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2024-12-10T05:24:48Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-version-not-used
+          note: 'This vulnerability was remediated in mattermost v7.x, Specifically, in versions 7.1.4, 7.2.1, 7.3.1, and 7.4.0. For more information, please refer to: https://www.clouddefense.ai/cve/2022/CVE-2022-4019'
```
luhring, line 541: >For more information, please refer to: https://www.clouddefense.ai/cve/2022/CVE-2022-4019

Just curious, why are we including this link? At first glance, it seems strange to use this vendor as an authoritative reference here

jamie-albert: @luhring this was a carry over from a different version of mattermost, I could see the perspective how using a third party vendor as a source of truth would raise concern. Consider the following however, as to why it might be valued as a **component of our audit trail**. Mattermost is difficult to trace CVE IDs and GHSA IDs back to their security releases for a couple reasons. They use [their own ID system seen here](https://mattermost.com/security-updates/) in their security releases where the data for the releases is just data pulled from a backend somewhere and rendered on to the page ie `<tr data-index="1"><td>MMSA-2024-00396</td>` meaning there is no static link associated with any given ID and since it renders this information dynamically you can not copy a given position or text on the webpage as what exists in index 1 for example and assume with any confidence that tomorrow it will be there as this will change when they add to their releases. Due to this, the [NVD/GHSA pages simply point you to the security releases page](https://mattermost.com/security-updates/) and wish you good luck. Mattermost also do not strongly correlate those IDs to the NVD or GHSA repo on their release page. Happy path best practice is to search keywords in the mattermost releases from description in GHSA, find a strong confidence match, take the username to the individual credited with assisting (if provided in description). Search Hackerone with both GHSA/CVE ID and username (if filed there, I have seen mattermost use this platform a lot and I have had success with this process) find conversation page about the CVE, verify information trail is solid and if there is a fix version then implement the fix and file an advisory anyway because there are no fix versions hosted anywhere but the mattermost security release page. Propose update to GHSA and NVD (leave it better than we found it mentality), and submit your fix PR into wolfi-dev/os or enterprise-pacakges and/or then send out advisory PR. Its a pretty arduous process. I don't inherently have an issue with adding a vendor as supporting evidence to our findings but imo it should not be the only source of information.

luhring: The link you're adding seems to be to some 3rd party product that's consuming data without citing their source.

I'd like to better understand the vulnerability match here. This used NVD data instead of GHSA data, which can lead to lower quality matching. Is it possibly a false positive? Should this be associated with https://github.com/mattermost/mattermost-plugin-playbooks instead of mattermost itself? Do we know what software content inside our APK might apply to this vulnerability?

jamie-albert: This CVE correlates to the following [hackerone page](https://hackerone.com/reports/1685979), which in turn can be traced by date and user to this entry on the security release page for mattermost: 

*MMSA-2022-00118 | Medium | <=7.1.3<=7.2.0<=7.3.0 | 2022-10-14 | v7.4.0v7.3.1v7.2.1v7.1.4 | (Denial of Service) Fixed an issue where an authenticated user could send multiple requests containing a large payload to the Update Playbooks API endpoint and crash the server. Thanks to vultza for contributing to this improvement under the Mattermost responsible disclosure policy.*

This security release was a part of the mattermost server security releases not the playbooks which makes sense as the denial of surface target is the server and the playbooks are the vector of attack.

I retain this to be a **false-positive-determination, vulnerable-code-version-not-used** due the above evidence and is surfaced by an overly broad CPE configuration that does not specify versions: `[cpe:2.3:a:mattermost:mattermost:-:*:*:*:*:*:*:*]` I have updated the advisory to reflect the more accurate info and will go back through to update the source mattermost advisory I pulled this from

luhring (approved):

PR #10112, opened 2024-12-17
fix(duplicate detection events): Remove duplicate detection events where the last event is the duplicate detection event

philroche:
I see no reason for these to be created as duplicates as the previous event was a fixed event or a pending-upstream-fix advisory


cpanato (approved):
will force merge

PR #11070, opened 2025-01-03
Samaba Advisory Update: 4.21.2

kbsteere:
Updated advisories for 8 CVEs with Samba.

### DIFF
--- a/samba.advisories.yaml
+++ b/samba.advisories.yaml
@@ -21,6 +21,10 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2025-01-03T20:49:25Z
+        type: fix-not-planned
+        data:
+          note: 'Security released not planned, some changes were added to address timeouts of MaxQueryDuration as referenced here: https://gitlab.com/samba-team/samba/-/merge_requests/2259 Future releases may receive more updates but not planned fix as of this time'
 
   - id: CGA-4jwm-758m-3ch2
     aliases:
@@ -39,6 +43,10 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2025-01-03T21:28:42Z
+        type: true-positive-determination
+        data:
+          note: 'Issue has been found in all 4.x versions of samba, Upstream has no plan to resolve.'
 
   - id: CGA-4r4x-rr4g-jr39
     aliases:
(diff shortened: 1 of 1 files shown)


samba.advisories.yaml
```diff
@@ -111,6 +131,10 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2025-01-03T19:51:58Z
+        type: true-positive-determination
+        data:
+          note: arbitrary Code Execution is limited to HP NonStop Servers with software H06.x through H06.23.00 and J06.x through J06.12.00, when Samba is used, allows remote authenticated users to execute arbitrary code via unknown vectors. Upstream maintains must address issue.
```
powersj, line 137: Should this be "Upstream maintenance must address the issue"? Or "Fix must be delivered by upstream"?

samba.advisories.yaml
```diff
@@ -93,6 +109,10 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2025-01-03T21:25:03Z
+        type: true-positive-determination
+        data:
+          note: 'Issue has been found in all 4.x versions of samba, no disscussion yet on planned fixed. '
```
powersj, line 115: How about "Upstream has no plan to resolve." or has chosen not to resolve? I'd like to be clear on who has not had the discussion.

samba.advisories.yaml
```diff
@@ -165,3 +193,7 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2025-01-03T21:34:10Z
+        type: true-positive-determination
+        data:
+          note: 'Issue has been found in all 4.x versions of samba, no disscussion yet on planned fixed. '
```
powersj, line 199: Similar comment as above around clarifying who had the discussion.

samba.advisories.yaml
```diff
@@ -129,6 +153,10 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2025-01-03T21:07:22Z
+        type: pending-upstream-fix
+        data:
+          note: 'PR is in open state and still pending changes. No timeline given when it will be completed. PR is 2 years old: https://gitlab.com/samba-team/samba/-/merge_requests/2778'
```
powersj, line 159: This is a really small fix, is this not something we would grab? If not, can we say why?

powersj (changes requested):
Thanks for grouping these up! I have a few small comments/questions that would help me understand the next steps of each. Thanks!

samba.advisories.yaml
kbsteere: Yeah that's better phrasing. I'll update

samba.advisories.yaml
kbsteere: So normally yes I think we would do a diff but the current PR state is not ready even though a PR has been created. There still is discussion around what to fix. More like a draft PR.

powersj: Thanks! What do you think about calling it an unreviewed PR to avoid getting asked why we didn't take the change?

kbsteere: I changed this to true-positive-determination, note: Fix must be delivered by upstream, PR is two years old: https://gitlab.com/samba-team/samba/-/merge_requests/2778'

powersj (approved):
Thank you for the fixes!

PR #11082, opened 2025-01-05
pending-upstream-fix advisory for xh package: GHSA-h97m-ww89-6jmq

mamccorm:
Raising `pending-upstream-fix` advisory for GHSA-h97m-ww89-6jmq, related to the `idna` crate. The xh package has multiple dependencies that rrquire different versions of idna.

Attempts at upgrading the oldest (to remediate this CVE), result in additional dependencies needing upgraded (such as cookie_store). Attempting to upgrade those results in similar dependency issues. Will require a fix from upstream.

### DIFF
--- a/xh.advisories.yaml
+++ b/xh.advisories.yaml
@@ -42,6 +42,14 @@ advisories:
             componentType: rust-crate
             componentLocation: /usr/bin/xh
             scanner: grype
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
thank you submitted a patch upstream https://github.com/ducaale/xh/pull/397 which should fix this in the new release.

PR #11227, opened 2025-01-09
Advisories for flyway-11.1.1

mjramirez-dev:
These CVEs come from `guava-31.1-jre.jar` and `netty-common-4.1.86.Final.jar`, which are not present in the upstream repo. They are transitive dependencies of the parent dependency `databricks-jdbc-2.6.40.jar`, which is closed source from databricks.


mjramirez-dev:
Related package PR: https://github.com/wolfi-dev/os/pull/39141

EyeCantCU (approved):

PR #11280, opened 2025-01-15
py3-docker/CVE-2024-36621/CVE-2024-36623

jamie-albert:
## 1. **CVE-2024-36621/CVE-2024-36623**
- **false-positive-determination:** 
    -  component-vulnerability-mismatch:
       The issue regarding disintegration/imaging v1.6.2 where the index of the scan function in scanner.go can go out of bounds has an open PR https://github.com/disintegration/imaging/issues/165 but no implanted fix yet. While the project has not been updated in a long time as seen here: https://github.com/wolfi-dev/advisories/pull/9300#issuecomment-2517650044 there is no definitive evidence that it is unsupported or archived and thus is
This is reflected in many other advisories, This affects the Docker Engine.(Moby) The package is a client api and does not have the vulnerability in question.

### DIFF
--- a/py3-docker.advisories.yaml
+++ b/py3-docker.advisories.yaml
@@ -113,6 +113,11 @@ advisories:
             componentType: apk
             componentLocation: /.PKGINFO
             scanner: grype
+      - timestamp: 2025-01-15T11:35:05Z
+        type: false-positive-determination
+        data:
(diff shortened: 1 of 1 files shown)


mamccorm (approved):
consistent with the other similar false-positive detections in this advisories file

PR #11351, opened 2025-01-23
glibc: add CVE-2025-0395 detect advisory

sil2100:
https://sourceware.org/git/?p=glibc.git;a=blame;f=advisories/GLIBC-SA-2025-0001;h=45f8b8f181fa024611452518ca0147ede1b88c47;hb=HEAD


sil2100:
This fails checks because the glibc PR, even though merged, did not result in the build getting uploaded (a transient error). Will have to re-push it as a no-change rebuild bump.

imjasonh (approved):

PR #11355, opened 2025-01-23
pgadmin4-17 package withdrawn from wolfi

Aditvil-Dev:
Package moved from wolfi to enterprise.
https://github.com/wolfi-dev/os/blob/main/withdrawn-packages.txt#L76


Aditvil-Dev:
We don't need to withdraw on move from os to enterprise.

philroche:
Package was added to wofi/os in error and was withdrawn. Removing the advisory now too

philroche (approved):

PR #11373, opened 2025-01-24
drop clamav-1.4 advisories

imjasonh:
https://github.com/wolfi-dev/os/pull/40474


kbsteere (approved):

imjasonh:
I think I'm going to keep these advisories since the package itself still exists, it just won't get any new versions.

PR #11390, opened 2025-01-28
Added false positive advisory events for solr 9.8.0 (GHSA-4p5m-gvpf-f3x5 and GHSA-68r2-fwcg-qpm8)

hbh7:
https://github.com/chainguard-dev/CVE-Dashboard/issues/18344
https://github.com/chainguard-dev/CVE-Dashboard/issues/18346

Pretty sure these are false positives as the package is already up to 9.8.0, which is the fixed version. 
```
🔎 Scanning "/tmp/packages.wolfi.dev-os-x86_64-solr-9.8.0-r0.apk2499216938"
├── 📄 /usr/share/java/solr/server/solr-webapp/webapp/WEB-INF/lib/solr-core-9.8.0-SNAPSHOT.jar
│       📦 solr-core 9.8.0-SNAPSHOT (java-archive)
│           Medium CVE-2024-52012 GHSA-4p5m-gvpf-f3x5 fixed in 9.8.0
│           High CVE-2025-24814 GHSA-68r2-fwcg-qpm8 fixed in 9.8.0
```

### DIFF
--- a/solr.advisories.yaml
+++ b/solr.advisories.yaml
@@ -182,6 +182,11 @@ advisories:
             componentType: java-archive
             componentLocation: /usr/share/java/solr/server/solr-webapp/webapp/WEB-INF/lib/solr-core-9.8.0-SNAPSHOT.jar
             scanner: grype
+      - timestamp: 2025-01-28T19:23:25Z
+        type: false-positive-determination
+        data:
+          type: component-vulnerability-mismatch
+          note: The scanner is not recognizing 9.8.0-SNAPSHOT as a version containing the fix, despite the latest 9.8.0 version being built as part of the package and containing the fix.
 
   - id: CGA-99pp-9g3p-rqqh
     aliases:
@@ -216,6 +221,11 @@ advisories:
             componentType: java-archive
             componentLocation: /usr/share/java/solr/server/solr-webapp/webapp/WEB-INF/lib/solr-core-9.8.0-SNAPSHOT.jar
             scanner: grype
+      - timestamp: 2025-01-28T19:27:51Z
+        type: false-positive-determination
+        data:
+          type: component-vulnerability-mismatch
+          note: The scanner is not recognizing 9.8.0-SNAPSHOT as a version containing the fix, despite the latest 9.8.0 version being built as part of the package and containing the fix.
 
   - id: CGA-cpx3-q2rm-vq9w
     aliases:


jamie-albert:
@hbh7 Please put the GHSA ID in the aPR title or the initial comment so it is easier to search for past remediations in the repo by that ID.

hbh7:
Apologies, you're right, I definitely should do that. Thanks for the note

jamie-albert:
@hbh7 So there are a couple serious fundamental issues here. With advisories, and **especially** ones we are marking as false-positive, we can not be "pretty sure" we need to be **absolutely 100% sure with evidence to back it up.** a 9.8.0-SNAPSHOT.jar does not inherently have the same contents as 9.8.0.jar The next step would be seeing where the SNAPSHOT aspect is coming from because we are supposed to be building from a full version release tag. It is likely coming from somewhere in the build as a variation and is being manipulated into the package after the fact. Regardless, you need to either:
1. Definitively prove that it contains the same fixes then provide that as evidence and/or investigate if there is any reason to be using a snapshot build
2. Show that it is a part of a larger naming convention discrepancy and point to evidence as to why the scanner is returning a component-vuln mismatch.

hbh7:
Thanks for your feedback. Here's some more context and evidence to boost confidence in the analysis:

From the scan, we see:
```
$ wolfictl scan --remote solr
...
🔎 Scanning "/tmp/packages.wolfi.dev-os-aarch64-solr-9.8.0-r0.apk4011824296"
...
├── 📄 /usr/share/java/solr/server/solr-webapp/webapp/WEB-INF/lib/solr-core-9.8.0-SNAPSHOT.jar
│       📦 solr-core 9.8.0-SNAPSHOT (java-archive)
│           Medium CVE-2024-52012 GHSA-4p5m-gvpf-f3x5 fixed in 9.8.0
│           High CVE-2025-24814 GHSA-68r2-fwcg-qpm8 fixed in 9.8.0
...
```

Based on https://github.com/apache/solr/blob/releases/solr/9.8.0/build.gradle, I can modify the `solr.yaml` build section as follows:
```
./gradlew assembleDist -Dversion.suffix=NOTASNAPSHOTANYMORE
```

This overwrites the default tag `SNAPSHOT` with a custom value so we can confirm that the resulting `solr-core-9.8.0-<SNAPSHOT?>.jar` file is being generated locally vs being downloaded. 

After running the `assembleDist` command, we can validate the resulting data: 
```
ls melange-out/usr/share/java/solr/server/solr-webapp/webapp/WEB-INF/lib
...
simplemagic-1.17.jar
snappy-java-1.1.10.5.jar
solr-api-9.8.0-NOTASNAPSHOTANYMORE.jar
solr-core-9.8.0-NOTASNAPSHOTANYMORE.jar
solr-solrj-9.8.0-NOTASNAPSHOTANYMORE.jar
solr-solrj-streaming-9.8.0-NOTASNAPSHOTANYMORE.jar
solr-solrj-zookeeper-9.8.0-NOTASNAPSHOTANYMORE.jar
spatial4j-0.8.jar
stax2-api-4.2.2.jar
...
```

Looking at several of the `build.gradle` files also appear to confirm the jar in question is being generated locally:
https://github.com/apache/solr/blob/releases/solr/9.8.0/solr/webapp/build.gradle
https://github.com/apache/solr/blob/releases/solr/9.8.0/solr/core/build.gradle

To me, this points towards the `SNAPSHOT` marker merely being a default value when building locally, rather than an indication that the fix may not be present. Perhaps we modify our build to set the default suffix to blank if possible? What are your thoughts @jamie-albert?

jamie-albert:
Awesome, figured it would be something along those lines. With limited bandwidth, we'll go down the path the PR is already suggesting. This would be a good candidate for hygiene work as it will continue to create tickets over time that are mismatched.

jamie-albert (approved):

PR #11889, opened 2025-02-03
python-3.13: add detection for CVE-2025-0938

ajayk:


luhring (commented):
Do we need this? Usually we don't add manual detection events because they don't communicate anything beyond the fact that we're aware of this. And it looks like we're imminently going to have a fixed event here anyway

powersj:
Closing as this was handled by automation: https://github.com/wolfi-dev/advisories/pull/11908/

PR #13094, opened 2025-02-13
advisory updates for CVE-2025-25193

kbsteere:


jamie-albert:
Summary from the GHSA page: An unsafe reading of environment file could potentially cause a denial of service in Netty.
When loaded on an Windows application, Netty attemps to load a file that does not exist. If an attacker creates such a large file, the Netty application crash.

jamie-albert (approved):

PR #15787, opened 2025-03-16
advisory entry added for CVE-2025-1767 to rancher-fleet package

developer-guy:


rancher-fleet.advisories.yaml
```diff
@@ -43,6 +43,10 @@ advisories:
             componentType: go-module
             componentLocation: /usr/bin/fleetagent
             scanner: grype
+      - timestamp: 2025-03-16T17:23:20Z
+        type: fix-not-planned
```
mamccorm, line 47: Did upstream note they weren't going to fix this? (and is there a link to something we can reference) if not we should set to pending-upstream-fix

developer-guy: so should I change it as pending-upstream-fix?

powersj: Please do, we only use [fix-not-planned](https://eng.inky.wtf/docs/teams/sustaining/maintainance/cves/advisory-data/#pending-upstream-fix) if no fixes will ever be released for the package. If upstream could still be working on a fix, then we should default to pending-upstream-fix.

When using pending-upstream-fix, it is also best to explain why we were unable to provide a fix. For example, did updating a dependency break something else, or maybe a fix isn't available yet, or it is a very complicated fix that would better landed by upstream.

Does that help clarify the difference?

OddBloke (approved):

PR #16698, opened 2025-04-02
remove OSV job from wolfi that runs on enterprise only

cpanato:


luhring (approved):
Was this just never used?

cpanato:
we used in the past, but then we decided to have just one feed for OSV

PR #16700, opened 2025-04-02
fix: Revert "Adding detection events for cassandra-4.1 (#13657)"

philroche:
This reverts commit 29ec54c96df77021f7bccc1b8cd3e9430ca404b1.

These cassandra-4.1 APKs in wolfi-dev/os were scanned in error and should not have resulted in new detection events.

cassandra-4.1 is now maintained in entperise-packages where these CVEs have been remediated.


powersj (approved):
Looked at CI 2 failures expected, +1

PR #16705, opened 2025-04-03
fix: Revert "Adding detection events for logstash (#13844)"

philroche:
This reverts commit 23562e897d63b190d1b09d3cdfca5b0e6e426ce7.

These logstash APKs in wolfi-dev/os were scanned in error and should not have resulted in new detection events.

logstash was renamed to logstash-8 and being maintained using that name. The rename occurred in https://github.com/wolfi-dev/os/pull/30436

### DIFF
--- a/logstash.advisories.yaml
+++ b/logstash.advisories.yaml
@@ -78,24 +78,6 @@ advisories:
         data:
           fixed-version: 8.12.2-r2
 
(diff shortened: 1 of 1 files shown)


philroche:
CI failure is expected as we are removing detections that were added in error

powersj (approved):

kwmonroe (approved):

PR #16708, opened 2025-04-03
fix: Revert "Adding detection events for rancher-agent-2.9 (#13897)"

philroche:
This reverts commit a58eebfea0e652db279c676baf4b7cc6c30a913e.

These rancher-agent-2.9 APKs in wolfi-dev/os were scanned in error and should not have resulted in new detection events.

rancher-agent-2.9 is now maintained in entperise-packages where these CVEs have been remediated.


philroche:
CI failure is expected as we are removing detections added in error

powersj (approved):

PR #16831, opened 2025-04-09
dotnet-bootstrap-9: replace 9.0.200-r0 with 9.0.2-r0

OddBloke:
The 9.0.200 version was used incorrectly: our 9.0.2 package contains the 9.0.200 SDK which contains these fixes.


powersj:
@philroche @OddBloke do we still need this?

OddBloke:
dotnet-bootstrap-9 should only be used to build dotnet-9, so these advisories aren't strictly necessary. If we run into issues in future, we can revisit this.

PR #18614, opened 2025-05-01
splunk-otel-collector/advisory update

kbsteere:
updating advisory for GHSA-6jgm-j7h2-2fqg


kwmonroe (dismissed):

powersj (changes requested):
I believe validate is failing because the detection is not filled out, I believe you need to state it was manual detection:

```
        data:
          type: manual
```

powersj (approved):

PR #19077, opened 2025-06-08
Zookeeper: version fix for CVE Data

AnmolVirdi:
- Zookeeper was accidentally released with wrong versioning format which led to release of `zookeeper-3.9.4.0` which was merely a `release-candidate`
- We are fixing `zookeeper` versioning for `9.x.x.x` to `9.x.x` since the last `.0`,`.1`,`.2` represent release candidates.


xnox (approved):
@philroche this will need a force merge.

PR #19238, opened 2025-06-11
celeborn: Add advisory

dnegreira:
Add advisory for GHSA-wf8f-6423-gfxg

The jackson-core dependency is brought in by the Hadoop transient dependency, upstream will have to release a version of Hadoop with a more recent version of jackson-core and celeborn will have to update it's Hadoop dependency in order to pull in a fixed version of jackson-core.

### DIFF
--- a/celeborn-0.5.advisories.yaml
+++ b/celeborn-0.5.advisories.yaml
@@ -131,6 +131,10 @@ advisories:
             componentType: java-archive
             componentLocation: /usr/share/java/celeborn/jars/hadoop-client-runtime-3.4.1.jar
             scanner: grype
+      - timestamp: 2025-06-11T10:47:00Z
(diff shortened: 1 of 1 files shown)


celeborn.advisories.yaml
```diff
@@ -0,0 +1,15 @@
+schema-version: 2.0.2
+
+package:
+  name: celeborn
```
bentasker, line 4: The package name seems to be `celeborn-0.5` rather than Celeborn

bentasker (changes requested):

celeborn.advisories.yaml
dnegreira: OOf, good catch. Thanks.

bentasker (approved):

PR #19367, opened 2025-06-12
kubeflow-katib/adv update

kbsteere:
fp for CVE-2024-47081


luhring:
> Hmm apk version not found in index?

https://github.com/wolfi-dev/os/pull/56011 hasn't landed yet, so the fixed version isn't in the index yet.

luhring (approved):

PR #19375, opened 2025-06-13
argo-cd/adv update older packages

kbsteere:
fix-not-planned for older argo-cd advisories. Escalation reference: https://github.com/chainguard-dev/customer-issues/issues/2305

### DIFF
--- a/argo-cd-2.10.advisories.yaml
+++ b/argo-cd-2.10.advisories.yaml
@@ -139,3 +139,25 @@ advisories:
         type: fixed
         data:
           fixed-version: 2.10.8-r0
+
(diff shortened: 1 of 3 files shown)


philroche:
> I see many other similar PRs that revert detection events, can we please check that they are also gone from https://images.chainguard.dev/security ? 

Wow. That is not something I would expect to happen. I will track this.

philroche:
@xnox moved to internal tracking issue https://github.com/chainguard-dev/internal-dev/issues/13862

philroche:
I will merge

philroche (approved):

PR #20498, opened 2025-06-23
drupal: update false-positive advisory

astrojerms:
Updates drupal-11 cve as a false-positive.

### DIFF
--- a/drupal-11.advisories.yaml
+++ b/drupal-11.advisories.yaml
@@ -9,6 +9,11 @@ advisories:
       - CVE-2012-2306
       - GHSA-g4rv-c45g-4j68
     events:
+      - timestamp: 2025-06-23T19:40:33Z
+        type: false-positive-determination
+        data:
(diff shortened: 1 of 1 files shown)


drupal-11.advisories.yaml
```diff
@@ -9,6 +9,11 @@ advisories:
       - CVE-2012-2306
       - GHSA-g4rv-c45g-4j68
     events:
+      - timestamp: 2025-06-23T19:40:33Z
+        type: false-positive-determination
+        data:
+          type: vulnerable-code-version-not-used
+          note: This vulnerability is a false positive and not present in recent versions of Drupal.
```
mamccorm, line 16: Can we include some evidence / reasoning to back up that claim? can he high level, but usually if it was confirmed by upstream maintainers, we'd like a GH issue, or write up our own why / how

kwmonroe (approved):

PR #21370, opened 2025-08-07
chore(kserve-modelmesh): Advisory for CGA-v2hv-cwjf-vh29

EyeCantCU:
Newer versions of netty provide a grpc implementation that is incompatible with kserve modelmesh

### DIFF
--- a/kserve-modelmesh.advisories.yaml
+++ b/kserve-modelmesh.advisories.yaml
@@ -204,7 +204,7 @@ advisories:
       - timestamp: 2025-08-07T02:50:38Z
         type: pending-upstream-fix
         data:
-          note: Netty 4.1.118.Final and newer provide:wq a gRPC implementation that is incompatible with kserve modelmesh.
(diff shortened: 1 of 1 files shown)


kserve-modelmesh.advisories.yaml
```diff
@@ -201,6 +201,10 @@ advisories:
             componentType: java-archive
             componentLocation: /opt/kserve/mmesh/lib/netty-handler-4.1.108.Final.jar
             scanner: grype
+      - timestamp: 2025-08-07T02:50:38Z
+        type: pending-upstream-fix
+        data:
+          note: Netty 4.1.118.Final and newer provide:wq a gRPC implementation that is incompatible with kserve modelmesh.
```
kranurag7, line 207:
```suggestion
          note: Netty 4.1.118.Final and newer provide a gRPC implementation that is incompatible with kserve modelmesh.
```

PR #22826, opened 2025-08-14
postgresql-17: mark CVE-2017-8806 as false positive

OddBloke:


AmberArcadia (approved):
> The Debian pg_ctlcluster, pg_createcluster, and pg_upgradecluster scripts, as distributed in the Debian postgresql-common package before 181+deb9u1 for PostgreSQL (and other packages related to Debian and Ubuntu), handled symbolic links insecurely, which could result in local denial of service by overwriting arbitrary files.

Agreed, we don't seem to distribute those.

PR #23248, opened 2025-09-01
chore: tidy up advisory file

bentasker:
25.1 is no longer the most recent version stream, so it has rotated out


efbar (dismissed):

bentasker:
Doh, you're right - what I should have done really is just to correct the version strings. Ta

philroche:
Why did we need this change? 

The latest version of 25.1 in os is 25.1.9-r1 so referencing 25.1.10-r0 doesn't same right. 

If anything I would remove the fixed event from these if they were only fixed after move to enterprise or in 25.2.

philroche (approved):

PR #23454, opened 2025-09-14
Nack CVE-2025-50817.

dlorenc:
It's not a real vulnerability, and the maintainers agree and have contested it.


kbsteere:
Adding reference to the upstream comments if anyone else is searching for it: https://github.com/PythonCharmers/python-future/issues/650#issuecomment-3252409616

